PR #7, opened 2016-09-29
Feature/api express

sylvainv:
The assignment was to use Loopback and Couchdb. I have made an implementation under the branch `feature/api-loopback`.

However after struggling some time with the API. I have come to the conclusion to try using express. Here are the reasons:
- Loopback is a very opinionated framework,
- I found it difficult to step out of the boundaries that are set for you, knowing how to program in Node.js I had found that frustrating,
- It is frustrating to have to use Loopback when I would know how to do without,
- All in all, the framework may be not bad but I think it is difficult to master in a short amount of time. There are just too many features.
- I found the documentation hard to navigate

More importantly, I am not extremely keen on using database abstraction, they make you avoid thinking about the database you are using, and unaware of its characteristics. Using express allows more flexibility for now and for future coding.

Sticking to CouchDB for this assignment, using `express` and [https://github.com/dscape/nano](a library to connect to CouchDB), allowed me to understand the database better, than if I had just sticked to an ORM or other abstractions library. This is I think a good thing and avoids making mistakes on your database choice, a pretty important one.

Several points I discovered about CouchDB:
- Database concepts seems to be a bit different, and dumping all documents into a single one may cause performance issues http://stackoverflow.com/questions/32504517/couchdb-update-multiple-databases 
- One can add validation on the server using design documents, but it is not possible to create default values on the server side for your document upon insertion,
- It seems there is a bug on the server and I could not find a way to throw bad requests (`throw(bad_request, 'Error message')` crashes the query)
- The documentation is quite complex to navigate and lack information

This version has the following feature: environment configuration, json web token authentication, CRUD operation on the user model.


xavierchow:
Thanks for your time and PRs.
Well, loopback is not only DB abstraction, you can focus on API layer with loopback remoting/model, or if you want to figure out the DB layer you can have a look at  [loopback-connector](https://docs.strongloop.com/display/public/LB/Connecting+models+to+data+sources).
I agree with some of your viewpoint, it's indeed complex and the documentation is not very exhaustive. I have no bias between loopback/express/koa or whatever but currently most of our projects are based on loopback, anyway I will check this PR as well.

index.js
```diff
@@ -0,0 +1,125 @@
...
+
+// AUTHENTICATION
+var authentication = function(model) {
+  return function(req, res, next) {
+    var authorizationheader = req.get('Authorization');
+    if (authorizationheader === undefined) {
+      res.send(401, 'Missing authorization header');
+    }
+
+    var tokens = authorizationheader.split('Bearer ');
```
xavierchow: Hmm, `split` might not be the best choice here, what if a customer header as `MyBearer xxxx`? Can you improve this with a regular express or whatever more solid?

index.js
```diff
@@ -0,0 +1,125 @@
...
+// PARSE
+var bodyParser = require('body-parser')
+app.use(bodyParser.json());
+
+// AUTHENTICATION
+var authentication = function(model) {
+  return function(req, res, next) {
+    var authorizationheader = req.get('Authorization');
+    if (authorizationheader === undefined) {
+      res.send(401, 'Missing authorization header');
```
sylvainv: Indeed. Fixed.

index.js
sylvainv: That's right, I use a capturing regex in my new commit. Although nobody should have something other than a bearer token in this implementation.

test/integration/user.js
```diff
@@ -0,0 +1,157 @@
...
+describe('Users', function () {
+  var app, token, docs;
+
+  before(function (done) {
+    docs = [];
+
+    app = request(wct.app);
+
+    // FIXME make sure app is started
+    setTimeout(function() {
```
sylvainv: I don't indeed. Fixed.

test/integration/user.js
```diff
@@ -0,0 +1,157 @@
...
+
+          data['address'].should.be.a('string');
+          data['description'].should.be.a('string');
+
+          data['created_at'].should.be.a('string');
+          wct.moment(data['created_at'], 'ISO').isValid().should.be.true;
+          var today = wct.moment().format('YYYY-MM-DD');
+          wct.moment(data['created_at']).format('YYYY-MM-DD').should.be.eq(today);
+
+          docs.push(data);
```
sylvainv: I have to say it does. it totally breaks the isolation rule. The correct implementation should use fixtures. That would be a better solution. I'll implement it.

xavierchow:
Thanks, that would be pretty much it for our evaluation.

JuhaS:
Closing old PR

PR #9, opened 2016-12-18
Done

Xaverius:
Hey as the title says I'm done, not that one is ever really done, but I feel I got a good working solution down.


modules/utils.js
```diff
@@ -0,0 +1,48 @@
...
+ * @param  {Object}       callback.options additional options
+ *
+ * @param  {Object}       options          additional options that will be handled by server
+ * @param  {Number}       options.status   status to send by server
+ * @param  {String}       options.location location to send by server
+ */
+function sendResponse(callback, options) {
+    return (err, result) => {
+        if (err) {
+            callback(err, void 1, options);
```
xavierchow: Why do you prefer `void 1` rather than `undefined` or `null`?

routes/userRoute.js
```diff
@@ -0,0 +1,50 @@
...
+
+function list(params, callback) {
+    callback({ message: 'not allowed' });
+}
+
+function get(params, callback) {
+    if (params.id) {
+        User.findOne({ _id: params.id }, utils.send(callback));
+    } else {
+        utils.send(callback);
```
xavierchow: I guess the callback is not called here, right? utils.send just returns another function.

server.js
```diff
@@ -0,0 +1,90 @@
...
+                    logger.trace('RESPONSE:', status, body);
+
+                    if (options.location) {
+                        res.setHeader('Location', url + options.location);
+                    }
+
+                    res.send(status, body);
+                }
+
+                next();
```
xavierchow: Can you tell me why `next()` is still needed after the `res.send()`?

xavierchow:
Hey, this is another Xavier from wiredcraft 😄 , thanks for you PR, just wanna challenge you a bit more,
is there anything can be applied to achieve a `really done`? (You don't need to finish it, just a TODO list up)

modules/utils.js
Xaverius: I could have used undefined here true, I just feel a bit uncomfortable writing `undefined` ever since I understood it could be redefined, although I guess on server side code that could never happen?
And `null` will eventually send a body with `null` whilst `undefined` will keep the body empty

routes/userRoute.js
Xaverius: It will be called, but the `else` path will never be called, but I have the habit of writing `else` when I write `if` just in case I didn't think of all possibilities and that could result in a server timeout.

server.js
Xaverius: Because of the restify library, it allows for multiple handlers for the same route and `next()` calls the next handler. You could add a general handler for logging, or for handling authentication headers for example.

Xaverius:
Hey Xavier :)

- For starters, I would implement an actual logger ;)
- I'd go over the model and validations more thoroughly, add actual messages so the caller will know what he's doing wrong. I'd also write unit tests with invalid values.
- I'd look up how the handle dates better, because I struggled a bit with that.
- I'd look up some scenario's how the server could crash and how to recover from that.
- In case this API would actually be used, I'd figure out how many calls it would handle and do some stress tests
- And last but not least (in case of an actual server) I'd make sure you needed to be authenticated

But that's out of the top of my head, I'm sure I could double this list in size.

xavierchow:
Thanks, that's pretty much enough for our evaluation

routes/userRoute.js
Xaverius: Actually it hit me during my sleep, I'm wrong, you're right it just returns a function that is never executed and I should have just called the callback.

PR #10, opened 2017-03-24
Interviewee/oliver.w

ole3021:
# Complete the test

* Create User data model
* Implementation the RESTFUL apis
* Implementation the login/logout feature and api
* Use MongoDB as persistent database
* Complete integration test

For more details and documentation about setup and run please check the `README` document on my branch.

For easier browse the code, you can check it out in my repo [test-backend](https://github.com/ole3021/test-backend/tree/interviewee/oliver.w)

Oliver.W @ 2017-03-24 20:07

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,70 +1,120 @@
 # Wiredcraft Back-end Developer Coding Test
 
-Make sure you read **all** of this document carefully, and follow the guidelines in it.
+## Initialization & Test
 
-## Background
+### Installation
 
-Build a restful api that could `get/create/update/delete` user data from a persistence database
+Clone the project to local drive, execute the following command to install dependencies
 
-### User Model
-
+```shell
+npm install
 ```
+
+Dependencies:
+
+- Node.js: V 7.7.4
+- MongoDB: V 3.2.0
+
+### Start Server
+
+Use the following command to start the server
+
+```shell
+npm run server
+```
+
+After server started, there will be expot the port 4020 on the machine as the backend service, to explorer the apis can visit the [API Explorer](http://0.0.0.0:4060/explorer), From the explorer there is a list of the exposed apis.
+
+### Test
+
+#### Lint the code
+
+```shell
+npm run lint
+```
+
+This will use [StandardJS](https://standardjs.com/) as the libiary lint the code styles in server folder.
+
+#### Post check on Code
+
+```shell
+npm run posttest
+```
+
+This will use [nsp](https://github.com/nodesecurity/nsp) to detect [known vulnerabilities](https://nodesecurity.io/advisories/) in the server code.
+
+>  In post check will include test `lint`
+
+### Integration Test
+
+1. Start server for test
+   In Test mode data will exist in memory instead persistent in database.
+
+   ```shell
+   npm run server-test
+   ```
+
+2. Start test work
+
+   ```shell
+   npm run test
+   ```
+
+   > In the test it will include `posttest`
+
+### Implementation
+
+#### User Model
+
+```javascript
 {
-  "id": "xxx",                  // user id(you can use uuid or the id provided by database, but need to be unique)
-  "name": "test",               // user name
-  "dob": "",                    // date of birth
-  "address": "",                // user address
-  "description": "",            // user description
-  "created_at": ""              // user created date
+"id": "String",         // user's id
+"user_name": "String",  // user's user name used during login
+"password": "String",   // encrypted user's password(hidden from user)
+"salt": "String",       // encrypt salt used to encrypt user's password(hidden from user)
+"name": "String",       // user's name
+"dob": "Date",          // user's date of birth
+"address": "String",    // user's address
+"created_at": "Date",   // record created time
+"modified_at": "Date"   // record modified time
 }
 ```
 
-### API
+The `user_name`, `password`, `salt` and `modified_at`are added beyond the original test requirement, to complete the login/logout task and complete the infos of tracking in future.
 
-```
-GET    /user/{id}                   - Get user by ID
-POST   /user/                       - To create a new user
-PUT    /user/{id}                   - To update an existing user with data
-DELETE /user/{id}                   - To delete a user from database
-```
+#### API
 
-## Getting started
+| Method | URI                     | Description                              |
+| :----: | :---------------------- | :--------------------------------------- |
+|  GET   | /api/users/{id}         | Get user by id                           |
+|  POST  | /api/users/             | Create new user                          |
+|  PUT   | /api/users/{id}         | Update User by id (user can only edit their slef data after login) |
+| DELETE | /api/users/{id}         | Delete user by id                        |
+|  POST  | /api/users/{id}/replace | Same to PUT /api/users/{id} (Loopback framework effect) |
+|  POST  | /api/users/login        | Login specific user with `user_name` and `password` |
+|  POST  | /api/users/logout       | Logout currently logined user(base on accessToken) |
 
-There's nothing here, we leave it to you to choose the build tool, code structure, framework, testing approach...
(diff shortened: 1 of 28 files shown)


xavierchow:
@ole3021  Thanks for your PR, I added a few comments and wanna challenge you a bit more:
Could you tell me what improvement can be done if your login&logout feature is gonna applied to a real production? (You don't have to implement it)

README.md
```diff
@@ -1,70 +1,120 @@
...
+To me, i think the implementation with the build-in User and AccessToken are kind of cheat for this test, so all the model in this test are based on loopback's `PersistedModel` which is the base model for loopback other models.
 
-## Q&A
+In this test during login part i use _Secure Salted Password Hashing_ to encrypted user's password in database, and use [uuid](https://github.com/kelektiv/node-uuid) to generate the logined user's token to verify, the verification has borrow some of the loopback lib code to make it can works in Loopback framework way.
 
-* Where should I send back the result when I'm done?
+Beside this i did the following to make the code be easier to expand and more friendly to end users.
 
-Fork this repo and send us a pull request when you think you are done. We don't have deadline for the task.
+* Create based model `WBase` to deal with general meta infos like `created_at` and `modified_at`
```
ole3021: Yes, the inheritance is the original idea that i used form the loopback framework, for Loopback's all model is based on the `PersistedModel`. 

For the question, if want to just simply implement the `created_at` field, it can declared in the `WUser`'s json file, and deal it with the hook `before_save` in `WUser` model.

The reason i make it like this, is based on my pervious experience, for the common and shared field in different model, keep it in the 'parent' can make things easier, but maybe i over-used it in this test.

package.json
```diff
@@ -0,0 +1,44 @@
...
+  "main": "server/server.js",
+  "scripts": {
+    "lint": "standard server/**/*.js",
+    "server": "nodemon . --exec babel-node --config ./configs/nodemon.json",
+    "server-test": "NODE_ENV=test babel-node .",
+    "test": "npm run posttest && ava -verbose",
+    "posttest": "npm run lint && nsp check"
+  },
+  "dependencies": {
+    "babel-cli": "^6.24.0",
```
ole3021: yes, it should be in devDependencies, and not in dependencies. 

This is a mistake i make, i imagine to use the `babel-node` in production, so it will be required in dependencies, but this my fault. 

* the `babel-node` should be not used in production environment (not in dependencies), cause it is pretty heavy with high memory useage.
* the correct way to do is build a compiled version of the Babel code into ES5 code, and execute the server with `node`

server/models/w-base.js
```diff
@@ -0,0 +1,26 @@
...
+module.exports = WBase => {
+  WBase.observe('before save', (ctx, next) => {
+    if (ctx.isNewInstance) {
+      // Add Created timeinfo
+      Object.assign(ctx.instance, {'created_at': new Date()})
+      next()
+    } else if (ctx.instance && ctx.instance.id) {
+      const { WUser } = app.models
+
+      WUser.findById(ctx.instance.id, (err, user) => {
```
ole3021: Yes, can alter the instance directly from the ctx.instance, in the ctx.instance it include the instance that in the request and going to save. 

The `findById` just make a waste on database request for this task. i didn't notice this/

xavierchow: no worries, 😄

server/models/w-user.js
```diff
@@ -0,0 +1,177 @@
...
+        WUser.findOne({
+          where: {
+            user_name: ctx.instance.user_name
+          }
+        }, (err, instance) => {
+          if (err) return next(ERRORS.MONGO_ERROR(err))
+
+          if (instance) {
+            // Duplicated user name
+            next(ERRORS.USER_NAME_BEEN_TAKEN)
```
ole3021: yes, this is could be have the TOCTOU problem, to solve the problem, should use the atomic method `findAndModify` of mongodb to make sure the find and create success on same time or fail on same time.

ole3021:
I will update and fix these problems with a new commit later today :)

README.md
makara: I guess @xavierchow was hinting this: https://www.npmjs.com/package/loopback-ds-timestamp-mixin

server/models/w-user.js
makara: What if I want to use a different DB?

makara:
You don't have to fix, just let us know your thoughts/opinions:

- You built a token model. What's the difference between that and the Express session?
- You built auth API with user model. How would you compare it with Passport auth? How would you compare it with cookie or JWT?

ole3021:
@makara For the questions

1. Both Token based and Session Base are ways to identify the users of request, and the difference are

   * The Token based auth is stateless than Session basesd auth, for the REST API purpose to do nonthing with users state, wo the token is more suitable for a RESTful API auth.
     * Session based auth will store some info/status on server and client, and server also have to manage these status.
     * Token based auth has no need to record the users state on server, client can be verified based on the token only.
   * Token based auth useall more flexable than Session based auth for API
     * Session are usually saved in browser cookie, which can cause problem where there have a  cross-doamin usage require with this kind authentication, for cookie is domain bassed.
     * Token can use cross-domain, cross-devices, enven in native app, which can make the API request with authentication easily.

2. There are many ways to do authentication, and my personal compare of theses are list below

   * My Implementation

     My implementation is baed on the original idea of Loopback authentication approach. beside it can be more flexable for future developemnt, for example, if i want to build role-level based access control system, it is more easier with this implementation ways, i can simply write some middleware and reuser the Loopback authentication menthod with my own authentication logic.

     > In my earlier developement, to build a more secure backend system, i build a role-level based authentication approach by myself, which used to auth users on each api request with different request method, and that can make sure the admin is more confidence on the access right management. 

     To implementation some customized like role-level baed, i found the custom implementation is easier to do compare to use Passport. and the code can be more readable. 

     The original idea of my implementaition is to get reuser of the most build-in features in Loopback , such as the authentication check approach, and just customize the smallest part to meet the requirements.


   * Passport Auth

     Passport is a middleware that make local, OAuth and Integrate other 3rd-part authentication(Google, WeChat) easier. For local auth it has username and password strategy, but it is more usually used with a session approach, another thing is the Passport is more a common authentication middleware, when i want to do more customized access right control, it can gets harder. For simple and quick implementation, the Passport is more suitable.

   * Cookie & JWT

     Both Cookie and JWT are saved more infos than just the token, but Cookie is domain related, and these info are more insecure, JWT is a more common way to add payload info into access token, and security enough, but the info that saved in token is not flexable to change unless to create another token. For example if i did some customized role based authentication and add that info in JWT, but after i update the role infos, the token still hold the old role info.

server/models/w-user.js
ole3021: For the TOCTOU problem i think it is more like a competition problem that happens in concurrent programming, but with DB, i think it should be the DB's responsibility to take over these problems, like the atomic method given in MongoDB.

But to avoid this problem in code logic, i think the following solutions
 * the way that solve this problem in concurrent programming can be refer,build a lock system to make sure the item is safe before any operation taken, and unlocked after operation success, but this also need to consider how to make the lock system work fast.

* A more easier way to avoid this kind of problem, is to use a queue if these logic can take the waiting time in the queue. a queue is more secure way deal with these kind of problem, all the operations are happens synchronously, there will be no worries about the competition.

README.md
ole3021: i checked this package, what it did is automatically inject the `created_at` and `modified_at` when the models load at the system start, and it also used the `observe` `before save` method to make this happen, this is a clean and simple way to make this happen instead i write it manually :smiley_cat:

makara:
Thanks. I think it's enough for us to evaluate. We'll get back to you soon.

PR #11, opened 2017-07-03
Test Backend - Alexis Lutun First version

Alutun:
Please refer to README for all the information. 
If you have any questions, remarks or suggestions, feel free to let me know

I'm not sure who to tag so I tag @makara since we meet each other.


test/test.js
```diff
@@ -0,0 +1,259 @@
...
+		.expect("Content-type",/json/)
+	    .expect(200)
+	    .end(function(err,res){
+	      res.status.should.equal(200);
+	      // console.log("res body is : " + util.inspect(res.body));
+	      res.body.name.should.equal("toto");
+	      res.body.description.should.equal("This user is a test");
+	      res.body.address.should.equal("Shanghai");
+	      res.body.should.have.property("id").which.is.a.Number();
+	      savedId = res.body.id;
```
xavierchow: The variable `savedId` here indicates the test codes are not independent and tight-coupled, could you figure out a way to decouple them?

Alutun: Hello @xavierchow let me rephrase me your comment to make sure I understand your point before I submit a commit .

You mean that because `saveId` is defined in the global scope, it's an issue because that link the different batch of test together because they rely on each other result. Is that right ?

xavierchow: Yes, e.g. you can't run mocha `only` with your `PATCH` test code cause it relies on `PUT`, but you don't really need to fix it, just tell me your solution.

xavierchow:
@Alutun I'm not sure how you handle the code style, but running `npm run lint` shows lots of error from my local environment.

test/test.js
Alutun: So if I want to make each test independent from each other I will do as follow. 

For example for DELETE endpoint Nested in the test 

1. I would first create an user json object which I will use to create a user with PUT.
2. Then Once the PUT is successful I will update my user object with the ID sent back inside my PUT answer. 
3. Then using that user json object I would do my test on the DELETE endpoint. 

So basically within each test, I would create the necessary HTTP resources needed to perform them independently. 

Let me know If my answer is clear enough.

Alutun:
@xavierchow Indeed you are right I forgot to run lint to check my syntax, my bad...  I'm in fact the first one usually make sure coding syntax are respected while working with other people. 
Do you want me to push a new commit fixing the syntax issue ?

test/test.js
xavierchow: Thanks, LGTM

xavierchow:
@Alutun Yes, please, FYI you can use eslint -fix to automatically fix most of issues, also make sure you are comfortable with [OTBS](http://wiki.c2.com/?OneTrueBraceStyle) and indent white space, different from C language it's kind of de facto style in node.js world.

Alutun:
@xavierchow Thanks for the tips, you totally spot my C language coding style. I got no more error when running `npm run lint`, using the loopback set of rules, let me know if it's ok on your local environment as well.

makara:
Thanks. That should be enough for us to take a decision. Closing.

PR #15, opened 2017-12-26
Backend Test Solution - Ahmed Gaber (Biga)

BigaDev:
Happy new year

Kindly find my solution for the backend test [here](https://github.com/BigaDev/test-backend).

I decide to not use Loopback since it has own User model with authentication and authorization functions, so I feel that it doesn't make sense to use it.I implemented the test using express.js and some NPM packages like mongoose (MongoDB), node-restful (generate CRUD APIs for user model), JWT (manage users tokens) and config (configuration management).


app/api/v1/user/service.js
```diff
@@ -0,0 +1,54 @@
...
+  verifyToken(res, token, next) {
+    jwt.verify(token, (err, userPayload) => {
+      if (err || userPayload === undefined) {
+        res.status(401).json({ success: false, message: 'invalid token' });
+      } else {
+        // valid token
+        User.findOne({
+          name: userPayload.username,
+        }, (userError, user) => {
+          if (userError) throw userError;
```
xavierchow: I'm not sure the Error thrown here will be caught somewhere else.

xavierchow:
Hi @BigaDev, this is Xavier from Wiredcraft, thanks for your PR and appreciated you built the authentication part, 
could you let me know what else can be improved for your authentication implementation if putting it to a real production?

BigaDev:
Hi @xavierchow, nice to meet you.
your question very interesting and I think there a lot of things can be done such as:

- setup a rate limit middleware for preventing authentication from brute force and DDOS attacks, we can use [express limiter](https://www.npmjs.com/package/express-limiter) for that purpose.
- generate predefined API key to ensure that who only owned this key can call authentication or even all APIs.

Additional points:

- secure our APIs using TLS (HTTPS).
- there is a good open source project implemented by strongloop called [API Gateway](https://github.com/strongloop/microgateway) they describe [here](https://strongloop.com/strongblog/node-js-loopback-api-gateway-sample-applications/) the main components which are very helpful to secure APIs, especially with microservices applications.

app/api/v1/user/service.js
BigaDev: Sorry, my mistake. I will fix it.

BigaDev: Fixed

PR #16, opened 2018-01-13
alexbharley test completed

AlexBHarley:


src/index.js
```diff
@@ -0,0 +1,87 @@
...
+ */
+userRouter.get('/:id', (req, res) => {
+  userService.get(req.params.id)
+    .then((user) => {
+      if (user) {
+        res.json(user)
+      } else {
+        res.status(200).send('User not found')
+      }
+    }).catch(e => res.status(400).send(e.message))
```
xavierchow: Could you tell me why you prefer 400 status code here?

src/db.js
```diff
@@ -0,0 +1,87 @@
...
+        } else {
+          result.id = result._id
+          delete result._id
+          resolve(result)
+        }
+      })
+    })
+  }
+
+  async update (collectionName, id, attributes) {
```
xavierchow: You built `create` and `get` without async/await, is there any reason for it?

src/index.js
```diff
@@ -0,0 +1,87 @@
...
+ * @api {post} /user/:id Update an existing user
+ * @apiName UpdateUser
+ * @apiGroup User
+ *
+ * @apiParam {String} [name]          the name of the user
+ * @apiParam {String} [address]       the users address
+ * @apiParam {String} [dob]           the users date of birth in the format YYYY-MM-DD
+ * @apiParam {String} [description]   the users description
+ */
+userRouter.post('/:id', (req, res) => {
```
xavierchow: Do you see any other options of http verb for the user updating?
Could you let me know why you chose to use `post ` here?

src/db.js
AlexBHarley: Not really, the `create` and `get` methods needed to do a transform on the data ie.

```
const item = r.ops[0]
item.id = r.insertedId
delete item._id
```

and it looked a bit cleaner inside the `new Promise`.

src/index.js
AlexBHarley: In a proper API I would distinguish between server (`5xx` error codes) (db queries failing, db connection failing, uncaught exceptions etc) and client (`4xx` error codes) failures. In this demo almost all that can go wrong is in data validation so I've just left it as 400.

src/index.js
AlexBHarley: Yes I could have used `PUT` or `PATCH`, some say `PATCH` should only update a single field so possibly `PUT` is more applicable (because we can update multiple fields at a time). 

Again it was just a time constraint thing, I didn't think too much about it.

PR #17, opened 2018-01-25
The test-backend is completed!

pivstone:
#### Brief:

-  Sails provider User RESTFull style API
-  heroku demo ✅
-  Dockerfile & docker-compose.yml ✅  
-  swagger docs ✅ 
-  login/sign up ✅  
-  simple unit test ✅ 
-  seneca ❎


xavierchow:
Hey @pivstone , thanks for your PR! Could you please indent your code with 2 space instead of tab for the ease of code review? I found some are mixed with tab and space.

api/controllers/AuthController.js
```diff
@@ -0,0 +1,49 @@
+/**
+ * AuthController
+ *
+ * @description :: Server-side logic for managing auths
+ * @help        :: See http://sailsjs.org/#!/documentation/concepts/Controllers
+ */
+
+module.exports = {
```
xavierchow: What else can be improved if putting your AuthController to a real production?
(You don't need to implement them, just some thoughts)

test/integration/helper.js
```diff
@@ -0,0 +1,28 @@
...
+  var agent = request.agent(sails.hooks.http.app);
+  agent
+    .post('/login')
+    .send({
+      username: 'test',
+      password: '123456'
+    })
+    .end(function(err, res) {
+      if (err) done(err);
+      done(agent);
```
xavierchow: done will be called twice if there is an error.

pivstone:
@xavierchow  Now, code is formatted by js-prettify.

api/controllers/AuthController.js
pivstone: I don't sure what kind change you want? Security or other ?

For security, the Authentication need other projection. 
- XSS, sails already provided via CSRF
- Throttling, limit the frequency of login for a same user(force password jack)
- Hash vulnerability, If user don’t existed, the don’t have a hash calculation.  It means the request response will more short than a login user existed.
- Log, wherever a user is login success or fail, we need a log to record who, when, where（ip）and what happened

for architecture,
- Token, if we practice micro-service, the AuthController need sign a  JWT or other kinds token, take over the token to the  API-gateway. share user and permission info to other service 
- Messege, broadcast a user login or logout event to the authentication topic. some other service need care about this type events. such as friend system or audit system

xavierchow:
Thanks @pivstone , that should be good enough for our evaluation, will come back to you soon,
closing.

PR #20, opened 2018-06-30
finished test

khier996:


maksim_khier/handlers/user_handlers.go
```diff
@@ -0,0 +1,106 @@
...
+  if err := c.ShouldBindJSON(&user); err != nil {
+    c.JSON(400, gin.H{"error": err.Error()})
+  } else {
+    if valid, errMes := user.Valid(); !valid { 
+      c.JSON(400, gin.H{"error": errMes})
+      return
+    } 
+    user.CreatedAt = time.Now()
+    if createErr := wiredDB.Create(&user).Error; createErr != nil {
+      c.JSON(400, gin.H{"error": createErr})
```
xavierchow: Why do you think `createErr` would be a 400(bad request)?

maksim_khier/handlers/user_handlers.go
```diff
@@ -0,0 +1,106 @@
...
+
+  var user models.User
+  if err := c.ShouldBindJSON(&user); err != nil {
+    c.JSON(400, gin.H{"error": err.Error()})
+    return
+  }  
+
+  user.ID = userId // make sure user_id is taken from url param, and not from payload
+  findUser := models.User{ID: userId}
+  if findData := wiredDB.First(&findUser); findData.Error != nil {
```
xavierchow: Why do you prefer to find the user first for updating? What if you update the model directly?

maksim_khier/handlers/user_handlers.go
khier996: Since there is already a User.Validate method that checks for the validity of a user, the checking of createErr is just an extra precaution. If some error happens on the database level, we need to inform the client. Now that I think about the exact status code, I am not sure 400 is the best choice. Maybe 500 would be better.

maksim_khier/handlers/user_handlers.go
khier996: We need to know if the user with the given id exists first. If the user doesn't exist, we need to inform the client. 
It is possible to directly update the model. This will create an sql query like:
UPDATE users SET name = 'new_name' WHERE id = 15
But if the user with id 15 doesn't exist, the query will not change anything. We will know that the number of affected rows is zero. But the number of affected rows is also zero when the new value is the same as the old value. Therefore, we cannot rely on this method to know for sure if the user exists or not.

maksim_khier/db/db.go
```diff
@@ -0,0 +1,54 @@
...
+  "flag"
+)
+
+var db *gorm.DB
+var dbErr error
+var logFile *os.File
+var logErr error
+
+func OpenDb() *gorm.DB {
+  setLogFile()
```
ChopperLee2011: I suggest setLogFile to be one step for app booting, not coupling to a DB operation.

maksim_khier/db/db.go
```diff
@@ -0,0 +1,54 @@
...
+  db.SetLogger(gorm.Logger{log.New(io.MultiWriter(logFile, os.Stdout), "\r\n", 0)})
+  return db
+}
+
+func setLogFile() {
+  if _, fileErr := os.Stat("log/log"); os.IsNotExist(fileErr) {
+    os.Mkdir("log", os.ModePerm)
+    logFile, logErr = os.Create("log/log")
+  } else {
+    logFile, logErr = os.OpenFile("log/log", os.O_RDWR|os.O_APPEND, 0666)
```
ChopperLee2011: how about closing log file, should it exist in the main script ?

maksim_khier/routes.go
```diff
@@ -0,0 +1,30 @@
...
+
+  api := router.Group("/api")
+  api.GET("/users", handlers.GetUsers)
+  api.GET("/user/:id", handlers.GetUser)
+  api.POST("/user", handlers.CreateUser)
+  api.PUT("/user/:id", handlers.UpdateUser)
+  api.DELETE("/user/:id", handlers.DeleteUser)
+}
+
+func setUpCors() {
```
ChopperLee2011: missing logging middleware: gin.Logger() i think

maksim_khier/main.go
```diff
@@ -0,0 +1,20 @@
...
+  "github.com/jinzhu/gorm"
+)
+
+var router *gin.Engine
+var wiredDB *gorm.DB
+
+func main() {
+  wiredDB = db.OpenDb()
+  wiredDB.AutoMigrate(&models.User{})
+  router = gin.Default()
```
ChopperLee2011: do you missing  `defer db.close()`, or you think it can be ignored?

maksim_khier/README.md
```diff
@@ -0,0 +1,75 @@
...
+```
+{
+  "name": "Boris",
+  "dob": "1950-01-01T12:00:00Z",
+  "address": "London",
+  "description": "UK foreign minister"
+}
+```
+
+###########################################################################
```
ChopperLee2011: - Provide a complete user auth (authentication/authorization/etc) strategy, such as OAuth.
I can not find this related API.

maksim_khier/db/db.go
khier996: i tried to do that, but couldn't find a convenient way to import db to all other packages

maksim_khier/db/db.go
khier996: fixed that in new commit

maksim_khier/README.md
khier996: added in new commit

PR #22, opened 2019-03-27
response - Justice Adeenze-Kangah

jakangah:
I have included a Response.md file explaining my methodology and thought process 
Let me know if you have any questions 
https://github.com/jakangah/test-backend/blob/develop/RESPONSE.md
Regards


src/services/http/patterns.js
```diff
@@ -0,0 +1,136 @@
...
+      bcrypt.genSalt(crypt.SALT_WORK_FACTOR, function (err, salt) {
+        bcrypt.hash(args.password, salt, function (err, hash) {
+          seneca.act('role:storage,cmd:insert', ({ type: 'administrator', input: { username: args.username, password: hash } }), function (err, res) {
+            console.log(res)
+            delete res.password
+            done(err, responseHandler(res))
+          })
+        })
+      })
+    } catch (err) {
```
xavierchow: do you think the try/catch can capture the errors when calling `bcrypt.genSalt` and `bcrypt.hash`?

src/services/http/connector.js
```diff
@@ -0,0 +1,203 @@
...
+         *       "status": 200
+         *
+         *
+         */
+      login: {
+        POST: true
+      },
+      /**
+         *
+         * @api {POST} /auth/signup adminSignup
```
xavierchow: what else do you think need to enhance the `/auth/signup` API if it's a real product API? (You don't have to implement it, just a discussion)

src/services/http/connector.js
```diff
@@ -0,0 +1,203 @@
...
+         *     HTTP/1.1 200 OK
+         *       {
+         *       "data": {
+         *           "_id": "5c99b8dce69885448104360d",
+         *           "name": "jaks",
+         *           "dob": "12/12/1990",
+         *           "address": "ashdjakhsdk",
+         *           "description": null
+         *       },
+         *       "status": 200
```
xavierchow: could you let me know the value of including the `status` in the response body considering we already have http statusCode?

src/services/http/patterns.js
```diff
@@ -0,0 +1,136 @@
...
+      } else {
+        done(err, responseHandler(message))
+      }
+    })
+  })
+
+  /**
+     * handles returning a user entity
+     */
+  seneca.add('role:user,cmd:load', async (msg, done) => {
```
xavierchow: it seems async is not needed here?

src/services/http/utils/passport.js
```diff
@@ -0,0 +1,38 @@
...
+const passport = require('passport')
+const JwtStrategy = require('passport-jwt').Strategy
+const ExtractJwt = require('passport-jwt').ExtractJwt
+const { jwt } = require('../../../config/vars')
+const { fetchAdmin } = require('../../storage/handler')
+
+/**
+ * Options for handling JWT requests
+ */
+let opts = {
```
xavierchow: I found you use a lot of `let` instead of `const`, could you let me know the reason?

jakangah: In most situations, I delegate configurations so separate files and use const to make sure they are unmutable, as such this would be better being pulled and safer(you can see examples of that in the index.js file).  The main config file is a const and as such I could easily pass that to the function instead

src/services/http/patterns.js
jakangah: Yes its not needed , you will note that in some of the other functions I omitted it , the reason is because when I started I had an idea where I could completely encapsulate the actions  , that way the functions could be resolved easily and maintained better instead of having them all in the patterns file , I would then be able to map them accordingly. Also in case I needed to use await this will enable me to go ahead and do that

src/services/http/connector.js
jakangah: My choice to include the status in the response comes from error handlin. Its easy to understand when the server returns an error of 400 stating an error, but if within the organization it might be possible that specific custom codes might require different actions. For example, we could have an error code 4443 that corresponds to a specific error caused by improper usage of a specific component of the api which would be more descriptive than a normal 400

src/services/http/patterns.js
jakangah: You’re right , would be better to async or isolate the process to ensure the right error is being called up the chain

src/services/http/connector.js
jakangah: rate limiting to prevent brute attacks (either with express or allowing a proxy to handle it ), reduce the expiration time for the tokens, revoking tokens ,  handling sessions

src/config/logger.js
```diff
@@ -0,0 +1,26 @@
+const winston = require('winston')
+
+const logger = winston.createLogger({
+  level: 'info',
+  format: winston.format.json(),
+  transports: [
+    new winston.transports.File({ filename: 'logs/error.log', level: 'error', handleExceptions: true }),
```
pivstone: If DevOps guys want to collect the app from 5 app instance, what should we do?

jakangah: I don’t understand your question, can you please clarify a bit

pivstone: Never mind,  It's not a good question.

jakangah: ah ok , no worries

pivstone:
Personal Feedback (Do not affect your task result):
- `http/connector.js` the file name is not very clear. I suggest use `http/route.js`. And other alternative name such as: `services/storage/mongodb/handler.js`, `http/auth/handler.js`. A good path/file name could help us to find the file easier.
- try to use [util.promiseify](https://nodejs.org/dist/latest-v8.x/docs/api/util.html#util_util_promisify_original) to wrap the callback function. Callback hell is language's flaw, but we can avoid it and improve our code readability.
- Do not full trust the unit test case.  I found I could post a empty json to sign up a account without error. Try to test your code manually.

jakangah:
Thanks for the feedback

xavierchow:
> Do not full trust the unit test case. I found I could post a empty json to sign up a account without error. Try to test your code manually.

I disagree, we should improve the unit test code to cover the edge case instead of testing it manually.

xavierchow:
closing.

PR #34, opened 2019-11-24
Solution from PaprikaZ

PaprikaZ:
New solution, read INTRODUCTION file before review.


xavierchow:
@PaprikaZ Thanks, it's good enough for our evaluation, closing.

PR #41, opened 2020-03-19
A General User Management Service

MiffyLiye:
## Documents

https://github.com/MiffyLiye/test-backend/blob/master/README.md


src/users/user.schema.js
```diff
@@ -0,0 +1,59 @@
...
+      type: String,
+      required: true,
+      validate: {
+        validator: function(v) {
+          return /^[a-zA-Z0-9]+$/.test(v);
+        },
+        message: props => `${props.value} is not a valid id!`,
+      },
+    },
+    name: { type: String, required: false },
```
ChopperLee2011: user.name is not required, which is unusual to me, do you have any concern about that?

src/auth/auth.service.js
```diff
@@ -0,0 +1,55 @@
...
+          this.configService.auth.hmacAlgorithm,
+          this.configService.auth.appSecret,
+        )
+        .update(timestamp)
+        .digest('hex')
+    ) {
+      return null;
+    }
+
+    if (
```
ChopperLee2011: why you have this design (not before or after AUTH_TIMESTAMP_TOLERANCE_IN_MS)?

src/users/users.controller.js
```diff
@@ -0,0 +1,123 @@
...
+        description,
+      });
+    } catch (err) {
+      if (err.name === 'ValidationError') {
+        throw new BadRequestException(err.message);
+      } else {
+        throw err;
+      }
+    }
+    return user.toJSON();
```
ChopperLee2011: do you know a mongoose method: lean(), what is the difference between lean(), instance.toJSON() and the plugin you are using for hidden some fields.

src/users/users.service.js
```diff
@@ -0,0 +1,55 @@
...
+    });
+    return user.save();
+  }
+
+  async getById(id) {
+    const user = await this.userModel.findOne({ id }).exec();
+    return user;
+  }
+
+  async update(id, { name, dob, address, description }) {
```
ChopperLee2011: if race condition, does this update method have issues?

README.md
```diff
@@ -1,73 +1,58 @@
...
+## Configuration
 
-> Where should I send back the result when I'm done?
+The system can be configured through environment variables
 
-Fork this repo and send us a pull request when you think it's ready for review. You don't have to finish everything prior and you can continue to work on it. We don't have a deadline for the task.
+| name | default | description |
+| ---- | ------- | ----------- |
+| APP_PORT | 3000 | the port this system listens to |
+| DATABASE_CONNECTION_URI | mongodb://localhost/users-service-development | the mongodb to use |
```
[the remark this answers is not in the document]
MiffyLiye: It's mentioned in the [development environments](https://github.com/MiffyLiye/test-backend/blob/master/README.md#development-environments) section that this system depends on `nodejs` and `mongodb`. 

If we also add `docker` to that section, then some npm scripts like `db:start`, `db:stop`, `db:migrate`, `db:reset` may be helpful for developers.

Sometimes I just use docker command to start local mongodb
```
docker run --name mongodb -p 127.0.0.1:27017:27017 -d mongo:4
```

But when I need to start many services like `zookeeper`, `kafka`, and `mongodb`, I use docker-compose to simplify local settings.

docs/api/users.md
```diff
@@ -0,0 +1,309 @@
...
+      "name": "MiffyLiye",
+      "dob": "2012-12-31",
+      "address": "China",
+      "description": "DEV",
+      "updatedAt": "2020-03-18T09:44:52.780Z",
+      "createdAt": "2020-03-18T09:44:52.780Z"
+    }
+```
+
+If the request failed, the response will contain the error message
```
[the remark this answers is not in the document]
MiffyLiye: When dob if not an valid date, we'll get response like
```
{
    "statusCode": 400,
    "message": "User validation failed: dob: 2012-02-31 is not a valid date!",
    "error": "Bad Request"
}
```

To distinguish user exists and field validation error, maybe we can use more specific 4xx status code, e.g. 409 for user exists. 
Or include a predefined error type enum string in the response.

src/users/user.schema.js
MiffyLiye: This system is designed to be used by another system, so I think less constraints are better.
If the `name` is required in business, the other system can add constraint in its own API, and send `name` to this system.

src/auth/auth.service.js
MiffyLiye: It's mentioned in the [authentication design decisions](https://github.com/MiffyLiye/test-backend/blob/master/docs/README.md#authentication-design-decisions) section, so that the service can decide when the auth headers should expire, and  the decrypted information will not pass authentication in the future since the timestamp is expired.

src/users/users.controller.js
MiffyLiye: I'm not familiar with `lean`. 
It seems that `lean` requires less CPU and memory, and will return pure data object including fields like `_id` and `__v`. 
The plugin I use to hide `_id` and `__v` only works with `toJSON` and `toObject`.
When we use `instance toJSON`, we will remove object methods including `instance.update` and `instance.remove`, and get pure data.

When saving machines is really important, I may try `lean` when it is a major bottleneck.
Currently I prefer not to use `lean`, to keep the concepts in this system fewer so that developers can work with it with less mental burden.

src/users/users.service.js
MiffyLiye: Yes, there may be some unexpected behaviors.

If we only take the `PUT` api or the `update` method into consideration:
Case 1. If the update requests come one by one, then the later request will overwrite the earlier request. 
Case 2. If two update requests come concurrently, then the changed fields will be merged, since mongoose only sends changed fields to the transaction command.

I think the behavior is not consistent, and may cause potential bugs.
One simple solution is to add a optimistic concurrency control (with a plugin like [mongoose-update-if-current](https://github.com/eoin-obrien/mongoose-update-if-current)), so that in Case 2 one of the request will be rejected. 
And the behavior will be simple and consistent.

If we take the whole `GET` and then `PUT` process into consideration:
1. Update request that finishes later will overwrite the request that finished earlier.

If the responsibility of user can be kept small, a `PUT` api is usually enough.
If we want to merge requests that each only changes part of the user, a `PATCH` api may be better.
If the responsibility of user grows to be bigger, we may need `PUT` api like `/users/:id/:subresource`.

src/users/users.controller.js
```diff
@@ -0,0 +1,123 @@
...
+
+@Controller('users')
+@Dependencies(UsersService)
+export class UsersController {
+  constructor(usersService) {
+    this.usersService = usersService;
+  }
+
+  @UseGuards(AuthGuard('HMAC'))
+  @Post('')
```
xavierchow: Why do you prefer `POST` to `PUT` for user creation?

test/api/users.spec.js
```diff
@@ -0,0 +1,375 @@
...
+      expect(user.description).toBe(description);
+      expect(user.createdAt).toBeTruthy();
+    },
+  );
+
+  it.each([
+    ['miffyliye', 'Miffy Liye', 'Alex'],
+    ['wangtao', 'Wang Tao', 'Bob'],
+  ])(
+    'should not create user when id is used by existing user',
```
xavierchow: If I understand correctly these tests are depending on the previous test result (https://github.com/Wiredcraft/test-backend/pull/41/files#diff-b192a391ed4193316fb05c0dfb46d100R25), right?
Do you see it as an issue?

test/api/users.spec.js
```diff
@@ -0,0 +1,375 @@
...
+
+      const createRes = await httpClient
+        .post('/users')
+        .set(authHeaders)
+        .send({
+          id,
+          dob: dateOfBirth,
+        });
+
+      expect(createRes.status).toBe(400);
```
xavierchow: I'm not sure the 400 error here is due to the invalid dob or the duplicated user ids(as you are reusing `miffyliye` and `wangtao`.

xavierchow:
@MiffyLiye thank you for your PR and reply.
Let me challenge you a bit more if you feel comfortable.
Regarding you consider what you built is a general user management service, let's say there is a consumer of your service which needs to import 10M(yes, million) users to your service; could you figure out a strategy of how to do it from both the consumer side and the service side?
Let me give you some requirements,
1. The consumer needs to know if the importing is successful or not, as some of the user ids might conflict with existing users in your system.
2. You have no control over when the importing will happen, i.e., you have to make sure your service is still able to serve for other consumers when the importing happens.
3. You can decide on how to implement the importing from the consumer.

I'm expecting a solution from you instead of a concrete implementation; surely, if you want to show something with the code, it would be welcomed as well.

src/users/users.controller.js
MiffyLiye: There are some business considerations.

Generally a user id may not uniquely identify a person in the real world if the consumer don't assign special meaning to it (e.g., passport number, employee id), so I think the consumer don't want to replace an existing user with same id, and `POST` is preferred.

If the consumer system use unique identifier of a person in the real world (e.g., passport number), then I think `PUT` (replace) may be better in this scenario.

test/api/users.spec.js
MiffyLiye: No, each test case runs independently. The database for test is dropped when the `AppManager` starts in `beforeEach`.

https://github.com/Wiredcraft/test-backend/pull/41/files#diff-b192a391ed4193316fb05c0dfb46d100R9

https://github.com/Wiredcraft/test-backend/pull/41/files#diff-b1654daea5b17178a70f3f91ce796f36R16

test/api/users.spec.js
MiffyLiye: Duplicated user error won't happen, because each test case runs independently. The database for test is dropped when the `AppManager` starts in `beforeEach`.

MiffyLiye:
@xavierchow 
The consumer that needs to import 10M users must be a very successful company. 

First, we need the infrastructure system to help scale the user management system automatically when there are high pressures in the database or in the stateless service.
We need to set `hash(id)` as the sharding key of the users in the database.
We need to start services behind a load balancer or load balancers.

If we don't add new API in the backend, we can provide a tool to help consumers do batch importing.
The consumer could do the following:
1. Set the `CONCURRENCY_DEGREE`, or use a default of 400. 
Higher `CONCURRENCY_DEGREE` will bring higher load to the system, but the importing will take shorter time. 
Lower `CONCURRENCY_DEGREE` will bring lower load to the system, but the importing will take longer time. 
If create 1 user takes 1 second, then with a `CONCURRENCY_DEGREE` of 400, it will take 6.94 hours to import 10M users. 
It will take less time if the network from the consumer to the system has high bandwidth and low latency.
2. The tool will take all the required settings from the consumer, and starts processing.
3. Partition the 10M users to `CONCURRENCY_DEGREE` groups, using `parseInt(hash(id).slice(-4), 16) % CONCURRENCY_DEGREE` as the group id.
4. Then the tool will start `CONCURRENCY_DEGREE` workers to process the grouped users, 1 worker for 1 group.
5. In each worker, it will send the request one by one, and record the result (success, user exists, unknown)
6. After all workers finished processing, the tool will merge the results from each group and generate a report of the processing result.

Then the consumer can decide what to do for failed user creations.

Also the batch importing can be implemented in the backend. 
A job scheduling subsystem is required. 
1. The consumer can create a job with `POST /jobs`, and add or batch add create user tasks with `POST /jobs/:jobId/tasks`.
2. Then the consumer can start the job with `PUT /jobs/:jobId/status {status: "start"}`.
3. The job scheduling subsystem will do what the consumer do in the previous design.
4. The consumer can see the job status with `GET /jobs/:jobId/status`.
5. When the job is finished, the consumer can see failed tasks with `GET /jobs/:jobId/tasks?some_query`

Then the consumer can decide what to do for failed user creations.

xavierchow:
@MiffyLiye Thanks for your answer, we are wrapping up the evaluation and will get back to you soon, closing this PR.

PR #2, opened 2015-08-30
A version without rest_framework

CCharlieLi:
1. Remove rest_framework folder from project
2. Add its installation step to README.md

PR #1, opened 2015-08-28
init commit

CCharlieLi:
1. Restful API function finished.
2. Demo deployed.
3. Test.

PR #45, opened 2020-06-09
A Koa2 RESTful API server

lalawow:
## Tech stack

- Nodejs and Koa2
- MongoDB and Mongoose
- RESTful API design
- Jest and supertest

## Docs
`./design_intro.md` describes how I designed this server.
`./how_to_run.md` describes how to run the server.


src/api/friendships.js
```diff
@@ -0,0 +1,106 @@
...
+      if (!ctx.params.userId) throw createError(400, "id in path is required.");
+
+      const req = {
+        userId: ctx.params.userId,
+        query: ctx.normalizedQuery || {},
+      };
+
+      const res = await this.listFriendshipsFromId(req, ctx);
+
+      if (!res.body) throw createError(500, "should have body in response");
```
superbogy, line 36: Do we need to return a 500 error?

src/services/auth.js
```diff
@@ -0,0 +1,135 @@
...
+  /**
+   * Logout
+   *
+   * @param {LogoutRequest} req logout request
+   * @param {import("koa").Context} ctx koa context
+   */
+  async logout(req, ctx) {
+    const { authorization } = req;
+    const access_token = authorization.split(" ")[1];
+    const sessionId = decodeJwtToken(access_token).session;
```
superbogy, line 128: Is it `decodeJwtToken ` can return `undefined` if `access_token` is invalid.

src/services/friendships.js
```diff
@@ -0,0 +1,108 @@
...
+    const { from, to } = ctx.params;
+
+    if (from === to) ctx.throw(400, "Wrong id!");
+    if (!isObjectId(from) || !isObjectId(to)) ctx.throw(400, "Wrong id!");
+    const users = await UserModel.find({ _id: { $in: [from, to] } });
+    if (users.length !== 2) ctx.throw(400, "Wrong id!");
+    const repeatFriendship = await FriendshipModel.findOne({ from, to });
+    if (repeatFriendship) ctx.throw(400, "Friendship already exists!");
+
+    const friendship = await FriendshipModel.create({ from, to });
```
superbogy, line 89: Can other users create a Friendship for `me`? 
Or is this API just for super admin?

src/services/friendships.js
```diff
@@ -0,0 +1,108 @@
...
+  /**
+   * Delete friendship by users id
+   *
+   * @param {DeleteFriendshipRequest} req deleteFriendship request
+   * @param {import("koa").Context} ctx koa context
+   */
+  async deleteFriendship(req, ctx) {
+    const { from, to } = ctx.params;
+    const friendship = await FriendshipModel.findOne({ from, to });
+    if (friendship) await friendship.delete();
```
superbogy, line 102: dangerous operation

src/services/users.js
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  /**
+   * Delete user by id
+   *
+   * @param {DeleteUserRequest} req deleteUser request
+   * @param {import("koa").Context} ctx koa context
+   */
+  async deleteUser(req, ctx) {
+    await ctx.state.user.delete();
```
superbogy, line 114: Dangerous operation

superbogy:
I am confusing about the API design, is it designed for admin? but the auth strategy looks like it is for normal users. Many updates do not verify the user's identity.
If I misunderstand pls point it out.

src/services/users.js
```diff
@@ -0,0 +1,119 @@
...
+   * Update user
+   *
+   * @param {UpdateUserRequest} req updateUser request
+   * @param {import("koa").Context} ctx koa context
+   * @returns {UpdateUserResponse} The user
+   */
+  async updateUser(req, ctx) {
+    const user = ctx.state.user;
+    const doc = pick(req.body, [
+      "name",
```
superbogy, line 97: You had made a unique name at `createUser` so here should not modify user's `name` without a check.

src/api/friendships.js
lalawow: The API user won't get the error message here except the status code 500. But we create error in this way can give backend log better explanation about this error incident.

src/services/auth.js
lalawow: In `./src/app.js` we used middlewares to ensure only valid access_token with valid session can reach this `logout` method. 
The `jwt` middleware verifies whether the token is jwt valid.
The  `sessionValid` middleware verifies whether the session is active.

src/job.js
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env node
+
+import { logger } from "./index";
+
+export function job() {
```
xavierchow, line 5: what is this function for?

src/lib/mongoose-helper/filter.js
```diff
@@ -0,0 +1,74 @@
...
+ * Build filter for list function
+ *
+ * @constructor
+ * @param {Object} raw raw filter
+ * @param {Schema} schema mongoose schema
+ * @returns {Object} normalized filter
+ */
+export default function build(raw, schema) {
+  const filter = Object.keys(raw).reduce((acc, key) => {
+    // TODO: 大部分逻辑移到 qn 里了，后面mongoose helper 需要从 qn 的标准query object 到 mongoose query 做转变
```
xavierchow, line 21: hmm, it this code snippet written by you or you borrowed it from somewhere?

src/services/users.js
```diff
@@ -0,0 +1,119 @@
...
+      throw ctx.throw(404, "user name not found.");
+    }
+    const repeatUser = await UserModel.findOne({ name });
+
+    if (repeatUser) {
+      if (ctx) throw ctx.throw(404, "user already exsists");
+      else return;
+    }
+    const user = await UserModel.create(body);
+    await PasswordModel.create({
```
xavierchow, line 69: As there are 2 creating operations it's possible to have inconsistent data around the user.id(e.g. the `PasswordModel.create` fails.
Do you have any idea to fix or mitigate the issue?
You don't need to update the code just put your ideas here.

lalawow:
Your concern has a very good point. The lack of verification design can
expose the system to some risks.
I haven't started to design the authorization rules yet, since I think it
depends on the users' needs and it's case by case.
We can put a "role" in User Model, and use "authorizationValid()"
middleware to do some verification.
Before that, sure, if we only allow users involved and admin to operate
those update/delete, that will be better practice for safety concern.

src/services/users.js
lalawow: Sure. I planed to do it, but forget to put it into the code.
Another thing, I will not allow `admin` to change name since we don't have a user `role` design yet.

src/job.js
lalawow: There's nothing there. But we could fill some jobs in it to run except the service, like prepare the database before runing the service.

src/lib/mongoose-helper/filter.js
lalawow: I borrowed `mongoose-helper` from @36node/mongoose-helper, and made some tiny modification to meet my design setting. It's very convient to use this to design `list` API for me.

src/services/users.js
lalawow: I didn't have concern here before, since mongoosedocument says model.create() returns a promise.

If we try to avoid this problem, I will try :-)
Method #1:
use `new Model()`, it should return user.id, too.
```ES6
    const user = new UserModel(body)
    await user.save()
    await PasswordModel.create({
      user: user.id,
      password: pwdEncrypt(DEFAULT_PWD),
    });
```

Method #2: 
We can put password into User Model as an attribute with`hideJSON: true`, and get rid of Password Model.
Then we only need create User once, without worry about the password creation here.

```ES6 
export const userSchema = new mongoose.Schema(
  {
    name: String,
    _password: { type: String, hideJSON: true }
  },
  {...}
);
```

And we write set and checkpassword method for password:
```ES6
class User {
  // other codes
set password(pwd) {
    this._password = encrypt(pwd)
  }

  checkPassword(pwd) {
    return verify(pwd,this._password)
  }
}
```

lalawow: `hideJSON` is one of property params of `mongoose-hidden` library which is a mongose schema plugin.

https://github.com/mblarsen/mongoose-hidden

xavierchow: Thanks for the ideas, but I don't think a promise will always be resolved, it could fail and throw, no?

> I didn't have concern here before, since mongoosedocument says model.create() returns a promise.

lalawow: Sure, when the database got trouble like too much work flood in, it could happen. I will keep that in my mind for the future design.

xavierchow:
Thanks all, this PR is good enough for our evaluation, closing

PR #47, opened 2020-06-12
An implementation in TypeScript

bsdelf:
Features

- Authentication based on JWT
- Structured & unstructured logging
- Follow and unfollow
- Nearby friends searching based on geographic coordinate
- Swagger documentation
- User soft deletion
- Route validation based on AJV
- CircleCI for lint & build & test

Features come from skeleton (barebone-js):

- Pino logging
- AVA unit test
- Sequelize migration
- Fastify web framework
- Stacked configurations
- Graceful kill

Development guide: [https://github.com/bsdelf/test-backend/blob/master/docs/guide.md](https://github.com/bsdelf/test-backend/blob/master/docs/guide.md)

barebone-js design: [https://github.com/bsdelf/barebone-js/blob/master/README.md](https://github.com/bsdelf/barebone-js/blob/master/README.md)

CircleCI build: [https://circleci.com/gh/bsdelf/test-backend](https://circleci.com/gh/bsdelf/test-backend)


MiffyLiye:
There are a lot of code, I will ask some high level questions here.

1. I found the `src/controllers` is like the service layer in other frameworks,
and `src/routes` like the controller layer in other frameworks.
How do you assign the responsibilities of `src/routes`, `src/controllers`, `src/models`.

2. How do you decide what to test in `test/routes`, `test/controllers`, and `test/models`?

3. The code contains log level, but it's not in the default config file, can we set the log level?
This code used many `logger.info`, and a few `logger.warn` and `logger.error`, how do you decide when to use these with different levels.

4. Why use `keyPrefix` in redis config, could you compare the pros and cons with other approaches?

bsdelf:
> There are a lot of code, I will ask some high level questions here.
> 
> 1. I found the `src/controllers` is like the service layer in other frameworks,
>    and `src/routes` like the controller layer in other frameworks.
>    How do you assign the responsibilities of `src/routes`, `src/controllers`, `src/models`.

In barebone-js, the responsibilities of these components are:

- routes
    1. authenticate
    2. validate inputs
    3. decode inputs
    4. invoke related components with inputs to serve the request
    5. build and validate outputs (rule out sensitive data)
- controllers: responsible for bussiness logic.
- models: responsible for data manipulation.

Unlike other framework, in barebone-js we treat "routes" as a thin and discardable part, which means they are there just for IO (and maybe cache), they won't play key roles. Actually you will find out following tasks are very easy to achieve based on current implementation:

1. Use express or koa to replace fastify.
2. Implement JSON-RPC style HTTP API for users.
3. Implement a websocket API for users.

> 2. How do you decide what to test in `test/routes`, `test/controllers`, and `test/models`?

- Test for routes should care about API behaviour.
- Test for controllers should be focused on business logic.
- Test for models should be focused on data itself.

In this PR, the user model and user link model is generated by `npm run cli:dev migration:create` based on sequelize model template, they are just data definitions, so there is no need to do unit test to ensure the correctness of sequelize library.

While the user session model is based on Redis, it is written from scratch, so unit test is necessary. 

Besides, the tests in this PR are mainly for illustration purpose, so it may cover 100% cases.

> 3. The code contains log level, but it's not in the default config file, can we set the log level?
>    This code used many `logger.info`, and a few `logger.warn` and `logger.error`, how do you decide when to use these with different levels.

Log level feature can be implemented by add a "level" to the config definition and apply it in the logger provider (feed level parameter directly to pino).

For level usage:

- Use error for events which should be caught by logging system and reported to developers.
- Use warn for suspicious events,  milestone events, it can be helpful for backtracing problems.
- Use info for development and informative purpose.


> 4. Why use `keyPrefix` in redis config, could you compare the pros and cons with other approaches?

`keyPrefix` is a feature in barebone-js's redis provider, it is introduced for unit test purpose.

In practice, I suggest use exclusive redis instance for each service.

But still, you can use `keyPrefix` to annotate project scope. For example, to save budget you decide to use a single redis instance for bookstore bussiness and online chat bussiness. In this case, `keyPrefix` can be helpful (I don't like this idea though).

bsdelf:
> 3. The code contains log level, but it's not in the default config file, can we set the log level?

Implemented in this commit: https://github.com/Wiredcraft/test-backend/pull/47/commits/463d270f21d67cc8da46a64c4788e432b495e2ef

Usage:

```yaml
providers:
  logger:
    level: "warn"
```

bsdelf:
> 1. I found the `src/controllers` is like the service layer in other frameworks,

BTW, in barebone-js, there is also a service layer: context & providers.

The "services" in barebone-js are just like Linux sysv/systemd services. Each service (provider) has its own defintion: dependencies, how to create, how to destroy. With all these information, the boostrap function will eventually execute [this line](https://github.com/bsdelf/barebone-js/blob/1e6c0528b8560887b3a433a0b854abe091228acb/src/context/context.ts#L22), the underlying implementation will use topological sorting to figure out the proper construction order and instantiate all the service instances and put it into context.

The best part is barebone-js's `context` object has well type hinting support, when you implement a new provider under "src/providers" folder, the type information will be automatically collected and injected into `context`, so VS Code can give you hints.

bsdelf:
>    This code used many `logger.info`, and a few `logger.warn` and `logger.error`, how do you decide when to use these with different levels.

One more thing, most of the time, we do not explicitly call `logger.error`, we just throw an predefined error (see [errors.ts](https://github.com/bsdelf/test-backend/blob/master/src/libraries/errors.ts)) which already has things like level, status code, error message encapsulated. Eventually, the error handler responsible to process the error, e.g. logging and deliver it to error tracking system.

MiffyLiye:
Thanks, now I can understand the PR better.

xavierchow:
@bsdelf thanks for you pull request, it's good enough for our evaluation, closing.
We will get back to you soon.

PR #52, opened 2020-08-13
Backend Test Submission - Obed Osei Frimpong

oseifrimpong:
Kindly find the URL for the documentation below. 

[Documentation](https://github.com/oseifrimpong/test-backend/blob/develop/docs/Readme.md)


xavierchow:
Thanks for the PR, closing.

docs/Readme.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Design
+
+### Authentication
+
+#### UML Sequence Diagram
+
+<p>Below is the UML sequence diagram for the user login flow. Assuming the user has already been created. </p>
+
+![uml](jwt-uml.png)
```
MiffyLiye, line 34: what tools are used to draw this UML?
how can we update the UML if the interaction changes in the future?

docs/Readme.md
```diff
@@ -0,0 +1,80 @@
...
+
+*The sequence diagram above provides a visual representation of the authentication/authorization flow.*
+
+## Logging
+
+  Winston is used as a logger for the application. There is a dedicated `/logs` folder to store the logs during runtime. Logging to the console has also been enabled. I want to be able to log all everything under the `debug`, `error` and `info` levels.
+
+## Testing
+
+Acceptance tests has been for all controllers and services and services have been implemented.
```
MiffyLiye, line 54: What does `acceptance tests` mean here?
Usually `acceptance tests` are written based on `acceptance criteria`,
and `acceptance criteria` are written from the viewpoint of end users. 

For internal controllers and services, which internal components act as the end users of each controller and service?

docs/Readme.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Setup
+Use the following steps to setup the application.
+
+- Install `Node 10 or above`
+
+- Create a ```.env``` file by duplicating ```.env.example``` file. Edit the .env variables to conform with the setup you desire.
```
MiffyLiye, line 61: if we want to use Replica Set or Sharded Cluster, how can we update the `.env` config.

src/__tests__/acceptance/user.controller.acceptance.ts
```diff
@@ -0,0 +1,159 @@
...
+        .expect(204);
+      expect(response.status).to.equal(204);
+    });
+  });
+
+  async function cleanDB() {
+    await userRepo.deleteAll();
+  }
+
+  async function createUser(data: object) {
```
MiffyLiye, line 145: In acceptance tests of the application from the end users' perspective, only application's web APIs can be used to test, the end users cannot call application's internal methods.

So in the acceptance tests of the controller, method of the controller layer is preferred over the internal method of userRepo.

src/index.ts
```diff
@@ -0,0 +1,41 @@
...
+  await app.boot();
+  await app.start();
+
+  const url = app.restServer.url;
+  console.log(`Server is running at ${url}`);
+
+  // configure Logs
+  app.configure(LoggingBindings.COMPONENT).to({
+    enableFluent: false,
+    enableHttpAccessLog: true,
```
MiffyLiye, line 17: I didn't find access log with config in `.env.example`,
How can I configure the app to see the access log?

src/controllers/authentication.controller.ts
```diff
@@ -0,0 +1,166 @@
...
+   * Extracts and hashes plain password
+   * Saves User details with hashed password
+   * Uses provided user details to generate
+   * JWT token.
+   * @param newUserRequest
+   * @returns token
+   */
+  @post('/signup', {
+    responses: {
+      '200': {
```
MiffyLiye, line 42: there are some error cases not documented here.

docs/Readme.md
oseifrimpong: - I used [draw.io](https://app.diagrams.net/).
- Ideally with draw.io, I can save the file in Google Drive and edit it later.
   
Using a tool like [Mermaid](https://mermaid-js.github.io/mermaid/#/) or any text format UML tool would have been a better approach. This makes it easier to maintain. The scripted text can be included within the documentation and a tool/extension can be used for rendering.

src/controllers/authentication.controller.ts
oseifrimpong: Yes, you are right! I had that in mind initially but I think it escaped my mind before submitting the PR.
Adding the following will the endpoint will help. I have added it to the code as well but because of the closed PR, changes are not reflected here. 

```js  '400': {
        description: 'Wrong Request body Structure',
        content: {
          'application/json': {
            schema: {
              type: 'object',
              properties: {
                error: {
                  type: 'object',
                  properties: {
                    statusCode: {
                      type: 'integer'
                    },
                    name: {
                      type: 'string'
                    },
                    message: {
                      type: 'string'
                    } 
                  }
                }
              }
            }
          }
        }
      },
      '422': {
        description: 'Invalid Request body',
        content: {
          'application/json': {
            schema: {
              type: 'object',
              properties: {
                error: {
                  type: 'object',
                  properties: {
                    statusCode: {
                      type: 'integer'
                    },
                    name: {
                      type: 'string'
                    },
                    message: {
                      type: 'string'
                    },
                    code: {
                      type: 'integer'
                    },
                    details: {
                      type: 'array',
                      items: {
                        type: 'object'
                      }
                    }
                  }
                }
              }
            }
          }
        }
      }

docs/Readme.md
oseifrimpong: We can update the `.env` by adding a URI like below:

`mongodb://mongo:27xxx,mongo_2:27xxx,mongo_1:27xxx/wc-test-db?replicaSet=rs0`

src/index.ts
oseifrimpong: Yes, I did not add configuration in the application to expose access logs. 
In other to achieve that, a global interceptor can be injected to handle the HTTP access logs. 

A custom-built interceptor can be used or we can simply make use of Winston's  _WINSTON_HTTP_ACCESS_LOGGER_ available in `@loopback/logging`. 

You can potentially also use [Morgan](https://www.npmjs.com/package/morgan)

src/__tests__/acceptance/user.controller.acceptance.ts
oseifrimpong: Yes, you are right.

docs/Readme.md
oseifrimpong: Yes, the acceptance test in this context is for any internal component that will be consuming this auth service. 
That internal component acting as an end-user can be a gateway or a client-side application[UI]. 

Also, the acceptance tests here are based on an assumed criterion of how the aforementioned end-user.

PR #53, opened 2020-10-04
Test Backend Submission - Jonathan Dai

agreatfool:
Hi Xavier,

I have finished the backend test, and made a pull request. Please have a look. And also document could be found here: [README.md](https://github.com/agreatfool/test-backend/blob/master/doc/README.md).

Best Regards,
Jonathan Dai


src/dao/UserLink.ts
```diff
@@ -0,0 +1,23 @@
+import {Redis} from "../database/Redis";
+import {genDbKey} from "../utility/Utility";
+
+export const KEY_USER_FOLLOWER_ID = "User:Follower:%ID%";
+export const KEY_USER_FOLLOWING_ID = "User:Following:%ID%";
+
+export const followUser = async (userId: string, targetId: string) => {
+  await Redis.get().sadd(genDbKey(KEY_USER_FOLLOWER_ID, {"%ID%": targetId}), userId);
```
xavierchow, line 8: what if the 1st `sadd` succeeds but the 2nd one fails somehow, how to mitigate the data inconsistency in some edge-cases?

src/dao/UserLink.ts
```diff
@@ -0,0 +1,23 @@
...
+  await Redis.get().sadd(genDbKey(KEY_USER_FOLLOWING_ID, {"%ID%": userId}), targetId);
+};
+
+export const unfollowUser = async (userId: string, targetId: string) => {
+  await Redis.get().srem(genDbKey(KEY_USER_FOLLOWER_ID, {"%ID%": targetId}), userId);
+  await Redis.get().srem(genDbKey(KEY_USER_FOLLOWING_ID, {"%ID%": userId}), targetId);
+};
+
+export const getFollowers = async (userId: string) => {
+  return await Redis.get().smembers(genDbKey(KEY_USER_FOLLOWER_ID, {"%ID%": userId}));
```
xavierchow, line 18: Is there a specific reason to use `return await`? https://eslint.org/docs/rules/no-return-await.

src/controller/UserGeo.ts
```diff
@@ -0,0 +1,142 @@
...
+    const result = [] as string[];
+    const followers = await UserLinkDao.getFollowers(ctx.params.id);
+    const nearbyIds = await UserGeoDao.getNearbyUserIds(ctx.params.id);
+
+    for (const [, id] of nearbyIds.entries()) {
+      if (id !== ctx.params.id // userId self would be returned by redis, so shall be filtered here
+        && followers.indexOf(id) !== -1) { // only user's "friends" (follower, since it's passive action) shall be displayed
+        result.push(id);
+      }
+    }
```
xavierchow, line 136: Let's assume here we need to support millions of Users with geolocation info and now the response of `getNearbyUserIds` is not fast enough, can you figure out a way to optimize it?

src/dao/UserLink.ts
agreatfool: One solution here is: we could use `transaction` of redis, like: 

``` typescript
await Redis.get().multi().sadd().sadd().exec();
```

And the transaction of redis is `NOT SAFE` for atomic operation. See [the official doc](https://redis.io/topics/transactions#errors-inside-a-transaction).

> It's important to note that even when a command fails, all the other commands in the queue are processed – Redis will not stop the processing of commands.

---

The other solution here is: lua script:

Since there is no rollback of official redis commands. If we need to keep these two commands all success or all rolled back when any error encountered, we could use self defined lua script to do it. See [official doc](https://redis.io/commands/eval).

src/dao/UserLink.ts
agreatfool: No. I just didn't notice `return await` would add an additional microtask. As the code above, there is no reason using `return await`. `return` directly is enough.

src/controller/UserGeo.ts
agreatfool: We could save user position into `block`.

Geohash has precision, low precision means a `block`:

block1

Use this low precision geohash value as key. Save user data into corresponding `block`.

When we need to retrieve nearby users. Use position of user to calculate which `block` user belongs to. Retrieve data in this block and calculate nearby users via geohash values. If the radius is bigger than one `block`, then also retrieving data of surrounding `block`s.

block2

xavierchow: Thanks, I see your idea, can you also give a time complexity with [big O notation](https://en.wikipedia.org/wiki/Time_complexity) and explanation correspondingly with the new solution?

agreatfool: Case 1: update user position:

1. remove old position: `ZREM` `O(log(N))`, N being the number of elements in the sorted set
2. add new position: `GEOADD` `O(log(N))`, N being the number of elements in the sorted set

Total complexity: `O(log(N))`, N being the number of elements in the sorted set

---

Case 2: get nearby users, and searching radius is inside a block

1. get nearby members inside the sorted set: `GEORADIUSBYMEMBER` `O(N+log(M))`, N is the number of elements inside the bounding box of the circular area delimited by center and radius and M is the number of items inside the sorted set
2. get user friends list: `SMEMBERS` `O(N)`, N is the set cardinality
3. compare fetched nearby users to friends list, to determine nearby friends: worst `O(N*M)` (as codes below), N is nearby users count fetched and M is the length of friends list

```typescript
for (const [, nearbyUserId] of nearbyUsers.entries()) {
	if (friendIds.indexOf(nearbyUserId) !== -1) {
		// ...
	}
}
```

Total complexity: `O((N+log(M))*K)`, N is the number of elements inside the bounding box of the circular area delimited by center and radius and M is the number of items inside the sorted set and K is the length of friends list

---

Case 3: get nearby users, and radius is bigger than a block(meanings shall calculate 9 blocks)

1. retrieve all members(userId) from 9 blocks(sorted set): `ZRANGE` `O(log(N)+M)`, with N being the number of elements in the sorted set and M the number of elements returned
2. retireve all position data of members from 9 blocks: `GEOPOS` `O(log(N))`, where N is the number of elements in the sorted set
	* since GEOPOS requires all member values as params, so step 1 is required
3. calculate the distence between members and target user to filter "nearby users": `some geohash lib function` `O(N)`, where N is the number of all memebers in 9 blocks
4. same as step 2 of previous case
5. same as step 3 of previous case

Total complexity: `O((log(N)+M)*K)`, N being the number of elements in the sorted set(9 blocks) and M the number of elements returned(9 blocks) and K is the length of friends list

MiffyLiye:
Hi Jonathan, I have some questions about the architecture and the test strategy.

1. In the code, there are folders like `controller`, `dao`, `model`, `database`, what is the responsibility of each layer?
Unlike traditional MVC framework, this architecture doesn't have service layer, and controller layer calls dao layer directly, 
Can you compare this approach with skinny controller (fat model) approach, and service layer approach?

2. In the test code, there are some server API tests, with test data prepared in a global `before` section with stubs.
Can you compare this approach with preparing test data in each test case (in `describe.before` or `describe.it` section) approach?
Can you compare the benefit & cost of using real Redis service in test cases without stubs?

3. About the test scenarios in test cases, most of them are calling one API and assert the response, but the API specifications are more than API responses, e.g., user created with `POST /user` should be retrieved from `GET /user/:id`. Can you extend or update the current test strategy to support this kind of specifications?

4. Similar to the 3rd question, many assertions in test cases are more about format, less assertions on business meaning, e.g. `GET /usergeo/nearby` I would expect the test case to prepare two users, one in the range, the other out of the range, when get nearby users, the assertion assets one user is in the response, and the other user is not in the response.
Can you extend or update the current test strategy to support assertions with more business meaning?

agreatfool:
Hi @MiffyLiye, Let me answer the questions one by one.

**Question 1:**

Functionality of each layer:

* controller: Handling incoming requests, validate params via Joi, and send verified request to service layer (in current test implementation, "service logic" solved directly in controller).
* database: Database handler of redis, utiltiy functions at all. This layer is a helper of database handling.
* model: Model object of business concept, like "user", "friend", "cart", "item", "shop", etc.
* dao: Data access objects, define functions to help the CRUD of model objects. In this layer, there is no business logic.

Let me explain why there is no service layer in codes. Since this test is a small test, there is almost no business logic here. So I omitted the service layer according to the rule "quick and dirty". In other projects of mine, including office projects and side projects, there are service layer existing.

For your question, I think skinny controller fat model (or service approach) is better than leaving logics all in the controller layer. Controller shall only accept requests and validate them, then move the valid requests into model layer, this could separate duties very clearly.

For the choice between skinny controller fat model / service approach. If business logic is not too complex, it's good to use fat model, just move domain logic into model layer directly. And if the project or system is very complex, it's good to use service approach (n-tiers arch), separate domain logic into services. And I prefer service approach in the daily work (If the project I can control), since I think this approach is more clear and flexible.

---

Let me explain a bit about the unit test part. As you know the test project is a bit simple and lack of real business logic. So my concept of unit test part is: Using correct skills/libraries and make simple unit test codes to show you guys how I do unit test in other projects. And as you can see, the logic of test case maybe a bit too simple as you mentioned in questions.

---

**Question 2:**

As I mentioned above, I thought about the logic part too simplistically. In real case, preparing global data and preparing data for single test case are all possilbe to happen. Difference:

* global data: For the scenario all test cases in one spec are all interrelated, we can create global data to let all the test cases use. For example:
	* create global data: users: [u1, u2, u3, ...], total 10 users
	* test case 1, getUsersAll: expect array length to be 10
	* test case 2, deleteUser: expect the returned array length of getUsersAll to be 9, after deleteUser action
* single case data: For the scenario all test cases in one spec are all doing their own logics, not interrelated. Then data shall be prepared before each test case by themself.

Using real Redis service in test:

* benefit:
	* could ensure the redis api usage is consistent with programmer. For example:
		* programmer think the api A of redis shall return object (actually this api would return number as result), and wrote the codes along with test cases, and he made some test data and redis stubs, as you can imagine, the case passed
		* if real redis is used, this test case could not pass
	* could test some redis API with logic inside, like GEOREDIUS
		* some edge case, programmer may think it's ok, but the result of redis is not
	* could test the case when redis is failed
		* test the reaction of app is expected or not when redis is down
* cost:
	* make the unit test more complex, since test framework have to ensure the redis facility available
	* commonly, real redis using in tests requires CI system, then unit test case could foucus on it's business logic

---

**Question 3 and 4:**

I have added two spec as Question 3 and Question 4 required. Please have a look at:

* [test/server/2 - user_business.spec.ts](https://github.com/agreatfool/test-backend/blob/master/test/server/2%20-%20user_business.spec.ts)
* [test/server/3 - lbs_business.spec.ts](https://github.com/agreatfool/test-backend/blob/master/test/server/3%20-%20lbs_business.spec.ts)

xavierchow:
Thanks all, this is good enough for our evaluation, closing the PR and we will get back to you soon @agreatfool .

PR #54, opened 2020-10-28
Backend Test Submission - Emmanuel Kyeyune

emmanuelnk:
Hey guys, 

Sorry it took a while. Been a hectic week. This is my implementation of the back-end test so far. I've done the authentication/authorization challenge and explained how I'd do the rest in the README.md.
The online demo is hosted at https://wc-backend-test.herokuapp.com/ and the routes can be tested with Swagger-UI at https://wc-backend-test.herokuapp.com/swagger-html

Let me know what you think or when you're ready to discuss. 

Thanks


MiffyLiye:
Hi, I'm Tao, I will review the code and ask some questions.

src/controller/user.ts
```diff
@@ -0,0 +1,214 @@
...
+import { safeCall, response } from '../utils/helpers'
+
+@tagsAll(['User'])
+export default class UserController {
+    @request('get', '/users')
+    @summary('Find all users')
+    @responses({
+        200: { description: 'success' },
+        400: { description: 'error' },
+        401: { description: 'token authorization error' },
```
MiffyLiye, line 16: when will this API return 400, I cannot find it in implementation code and test code.

src/controller/user.ts
```diff
@@ -0,0 +1,214 @@
...
+        response(context, 204)
+    }
+
+    @request('delete', '/testusers')
+    @summary('Delete all test users')
+    @responses({
+        204: { description: 'all users deleted successfully' },
+        401: { description: 'token authorization error' },
+    })
+    public static async deleteTestUsers(context: BaseContext): Promise<void> {
```
MiffyLiye, line 200: who will use this API, and how can we prevent other users from using it.

test/api/user/post.users.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+    it('should create a new user without address or description', async () => {
+        const res = await request(testServer).post(endpoint.route).send({
+            name: 'Mina',
+            email: 'mina@test.com',
+            dob: '1996-05-29',
+            password: 'AAaa@@88$$99',
+        })
+
+        expect(res.status).to.be.equal(201)
+        expect(res.body).to.be.jsonSchema(userSchema)
```
MiffyLiye, line 117: can we assert the result from a more business oriented perspective, not only verify the syntax of the response, but also verify that the user is saved in the system that can be retrieved later?

test/api/user/get.users.id.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+
+        const res = await request(testServer)
+            .get(endpoint.route.replace(':id', 'some_bad_id'))
+            .set('Authorization', 'bearer ' + token)
+
+        expect(res.status).to.be.equal(400)
+    })
+
+    it('should get a user by id', async () => {
+        const users = await generateDatabaseUsers(database, 1)
```
MiffyLiye, line 48: here the user is created with direct database operations, another way to prepare data is to use the `POST /users` API, can you compare the pros & cons of these two approaches?

haishanh (commented):
Thanks for the PR.

I've left some nitpicky comments, you may safely ignore them.

One thing that confuses me is:

Why you don't return the refresh token to the user in the login handler `src/controller/auth.loginUser`? In the refresh token handler `src/controller/auth.refreshToken` you are allowing a user to refresh it's access token with an expired access token. Assuming config `jwt.accessTokenLife` is `15m` and `jwt.refreshTokenLife` is `24h`, then the effective access token "life" actually is 24h if the user is not logging out itself.

src/controller/user.ts
emmanuelnk: No it does not return `400`. `400` should not be among the responses for this endpoint in Swagger docs. An error on my part.

src/controller/user.ts
emmanuelnk: For the purposes of this back-end test, I exposed this endpoint to all authenticated test users to allow for testing on the online build. In actual production, I would not expose this kind of sensitive endpoint that can remove all users from the database. However, one has to be authenticated to use it for the test as well. 

In the rare scenario where such an endpoint actually is needed in production, I would restrict it to administrators and protect it with several layers of authentication:
1. Auth token to be authorize user on the server
2.  Check for user role (e.g. Super Admin, Admin, Mod, User, Subscriber). Only one as Super Admin can execute this action.
3. Possibly add another endpoint for Two-Factor Authentication and require this authentication to permit this sensitive admin action.

test/api/user/post.users.spec.ts
emmanuelnk: Yes, this is good practice. Asserting the syntax only is no assurance that the document is saved in the database. I missed implementing this assertion here. 
I actually do this for the `delete /user/:id` endpoint. I assert that that the document does not exist in the database. However, it should also be done for endpoints that create and update resources such as this to ensure the resource was actually saved/updated in the database I can update the test code.

test/api/user/get.users.id.spec.ts
emmanuelnk: **1. Creating the test user with direct database operations**
 **Pros:**
 - In the situation in which the `POST /user/:id` endpoint is broken or has an error, it would cause the other tests that depend on this endpoint to fail. Tests should ideally be as independent as possible to ensure that external factors that are not part of the test do not affect the results of the test.
 - Creating the user with direct database operation allows you to insert data that you may not be able to insert with `POST /user/:id` e.g. fields such as `createdAt`, `updatedAt`, `refreshToken`. This gives you greater control of the test since you can insert different kinds of data to test different test cases easily.

**Cons:**

- Direct database operation can fail because of an external factor such as the database connection failing or hanging. However this would have the same effect on using an endpoint to create test users.

**2. Creating a test user with API endpoint**
**Pros:**
- I can only think of one scenario where one may want to do this: In a _multi-step testing workflow_ that has several requests – This is where a series of requests are used to simulate the usage of the API e.g. Register/Login flow. 
For example, a POST request is executed and it creates a user and returns a token in its response. A user is then authenticated with this token on the server. Then we use a PATCH endpoint to update the user, and we again invoke a GET request to validate the new data. Finally, we DELETE that user and use GET again to verify it no longer exists.
This approach can be used to catch certain errors/bugs/usability issues that independent endpoint tests (with direct database creation of resources) may not catch.

**Cons:** 
- As detailed in the Pros of direct database connection.

**Conclusion**

The first approach is preferred but ideally both can be implemented. Independent integration API tests with direct database connection for creating resources run first and when they pass, a multi-step testing workflow can be run against the API e.g A tool such as [Postman Newman ](https://learning.postman.com/docs/running-collections/using-newman-cli/command-line-integration-with-newman/)test runner can be used for this.

emmanuelnk:
> Thanks for the PR.
> 
> I've left some nitpicky comments, you may safely ignore them.
> 
> One thing that confuses me is:
> 
> Why you don't return the refresh token to the user in the login handler `src/controller/auth.loginUser`? In the refresh token handler `src/controller/auth.refreshToken` you are allowing a user to refresh it's access token with an expired access token. Assuming config `jwt.accessTokenLife` is `15m` and `jwt.refreshTokenLife` is `24h`, then the effective access token "life" actually is 24h if the user is not logging out itself.

No worries, about the nitpicky comments, I love feedback. I'll address them too. 😁 

Regarding the refreshToken, I'll do my best to explain.

> Why you don't return the refresh token to the user in the login handler `src/controller/auth.loginUser`?

The short answer is because its not secure. [Section 8 ](https://tools.ietf.org/html/draft-ietf-oauth-browser-based-apps-00#section-8)of OAuth 2.0 for Browser-Based Apps states:

> Authorization servers SHOULD NOT issue refresh tokens to browser-based applications.

In browser-apps, storing the jwt `refreshToken` on the client exposes them to abuse/hijacking. A malicious user can hijack the `refreshToken` and use it to re-issue valid `accessTokens` because in a realistic scenario, the `jwt.refreshTokenLife` would be much longer than `24h`.  In mobile apps the `refreshToken` can be stored securely in the client phone and as such these concerns are not as serious. The server can issue the `refreshToken` to the app.

To allow secure sliding-sessions in browser apps, the `refreshToken` is saved in the database (or server).  It serves these purposes:
- Used to keep track of how long the sliding-session can last (till expiry date of the refresh token)
- Invalidates a users sliding-session in the case of security concerns by removing the refreshToken from the database
- Since this is a browser app, for security reasons and to meet [Section 8 ](https://tools.ietf.org/html/draft-ietf-oauth-browser-based-apps-00#section-8)of the OAuth 2.0 for Browser-Based Apps spec.

> then the effective access token "life" actually is 24h if the user is not logging out itself.

Yes, as intended. This is the sliding-session. [Auth0 ](https://auth0.com/blog/refresh-tokens-what-are-they-and-when-to-use-them/)defines the sliding session as:

> Sliding-sessions are sessions that expire after a period of inactivity. As you can imagine, this is easily implemented using access tokens and refresh tokens. When a user performs an action, a new access token is issued. If the user uses an expired access token, the session is considered inactive and a new access token is required. Whether this token can be obtained with a refresh token or a new authentication round is required is defined by the requirements of the development team.

After `24` hours or whenever the `refreshToken` expires/removed, the user is not allowed authenticated access on the server until they log in again (via `src/controller/auth.loginUser`) and get a new `accessToken` (while a new `refreshToken` is generated for them and stored in the user collection)

I hope this clears up any confusion.

package.json
```diff
@@ -0,0 +1,117 @@
...
+  "version": "1.0.0",
+  "description": "RESTful API using NodeJS, KOA, Typescript and TypeORM (Mongo). Middleware uses JWT, CORS, Winston Logger and Swagger API docs",
+  "main": "dist/server.js",
+  "scripts": {
+    "watch-server": "nodemon --watch src -e ts,tsx --exec ts-node src/server.ts",
+    "build-ts": "tsc",
+    "copy-static-assets": "ts-node copyStaticAssets.ts",
+    "serve": "node dist/server.js",
+    "build": "npm run lint && npm run build-ts && npm run copy-static-assets",
+    "start": "npm run serve",
```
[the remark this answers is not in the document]
emmanuelnk: Yes this is true. An oversight on my part.

src/controller/user.ts
```diff
@@ -0,0 +1,212 @@
...
+
+        if (validationErrors) return response(context, 400, validationErrors)
+
+        const call = await safeCall(userRepository.findOne(context.params.id))
+
+        if (call.err) return response(context, 500, call.err)
+
+        if (!call.res) return response(context, 400, 'user_not_found')
+
+        response(context, 200, call.res.toJSON())
```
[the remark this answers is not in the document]
emmanuelnk: Actually this is misnaming issue on my part. I should have called the function `clean()` ,`safe()` or `lean()`. All its really doing is removing protected/sensitive fields. 
```ts
    toJSON() {
        return _.omit(this, ['password', 'refreshToken'])
    }
```

src/utils/conn.ts
```diff
@@ -2,7 +2,7 @@ import { Connection, createConnection } from 'typeorm'
 import { Tedis } from 'tedis'
 import { config, Config } from './config'
 
-let conn: Connection | null = null
+let conn: Connection | undefined
```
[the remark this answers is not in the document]
emmanuelnk: Yes you are right. Unfortunately this is left over from when I was testing exporting a global database connection. It is does not need to be declared outside the function.

The function should be and `conn` doesn't need to be defined:
```ts
export async function setupConnection(url: string, drop: boolean = false): Promise<Connection> {
    return createConnection({
        type: 'mongodb',
        url,    
        useNewUrlParser: true,      
        ssl: config.dbsslconn,
        authSource: 'admin',
        keepAlive: 600000,
        w: 'majority',
        entities: config.dbEntitiesPath,
        synchronize: true,
        useUnifiedTopology: true,
        dropSchema: drop,
        logging: false,
    })
}
```

src/controller/auth.ts
```diff
@@ -0,0 +1,144 @@
...
+
+        if (missingProperties.length > 0) return response(context, 400, `missing_fields: ${missingProperties.join(', ')}`)
+
+        const userRepository: Repository<User> = getManager().getRepository(User)
+        let user: User = new User()
+        let call = await safeCall(userRepository.findOne({ where: { email: context.request.body.email } }))
+
+        if(call.err) return response(context, 500, call.err)
+
+        if(!call.res) return response(context, 400, 'user_not_found')
```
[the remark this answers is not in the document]
emmanuelnk: `safeCall` is used to handle 500 type errors caused by bad database connections etc. When `call.err` is defined, it means the server should respond with `500 Server Error`. If everything went well, then `call.err` is **not** defined and we can continue to handle `call.res` which can also be `undefined` or have the desired value returned from the async function. I believe I would still require two `if-statements` and thus keeping the number of branches the same? I'll try it out

test/api/user/get.users.id.spec.ts
MiffyLiye: current test case is depending on a test helper using the internal implementation details, 
how can we keep the data consistent as data created in normal execution interface of the same layer? 
it's likely that we will maintain the same implementation in two places, one in the `POST /users API`, one in the test helper.

can you explain more about test independence? 
what is the unit to test? what is the boundary of the unit that should be independent?
if each test case is a unit, then do not depend on test data in other test cases,
if each interface is a unit, then do not depend on other interface of the same module/layer, 
if each module/layer is a unit, then do not depend on internal implementation details in other modules/layers,
etc.

emmanuelnk: > How can we keep the data consistent as data created in normal execution interface of the same layer

- This could be accomplished with the insertion of captured state fixture documents that capture the correct state of documents in the database _as they should be_ after they have been _correctly_ created/modified by the API. This is what the `generateDbUsers()` function I use in the test cases does. 
- This can also be done by importing the logic of resource creation (e.g `user.createUser()` from the user controller) into the `generateDbUsers()` function to create test users to ensure consistent implementation.

> can you explain more about test independence?
> what is the unit to test? what is the boundary of the unit that should be independent?

With regards to these API tests, each test case should be independent of data from other test cases. These endpoint test cases can therefore be regarded as units in the sense that they should not depend on test data from other test cases.  

The test helper that generates test data for some of these test cases does not depend on data from other test cases. The test cases do however depend on data in another layer (database layer). 
e.g. 
- a test to `GET users/:id` assumes the user already exists in the database, and thus a test helper inserts this user in the correct state before the test runs.
- a test to `PATCH users/:id` assumes the user already exists in the database, and thus a test helper inserts this user in the correct state before the test runs.

Therefore, in this situation, what I mean by test independence is independence from data generated by other test cases (other endpoint tests).

src/controller/user.ts
```diff
@@ -0,0 +1,214 @@
...
+    @summary('Update a user')
+    @path({ id: { type: 'string', required: true, description: 'id of the user to update' }, data: { type: 'object', items: {} } })
+    @body(_.omit(userSchema.properties, ['id', 'createdAt', 'updatedAt', 'following']))
+    @responses({
+        200: { description: 'user updated successfully' },
+        400: { description: 'user not found, user already exists, validation errors, user already exists' },
+        401: { description: 'token authorization error' },
+        403: { description: 'not authorized to perform this action' },
+    })
+    public static async updateUser(context: BaseContext): Promise<void> {
```
haishanh, line 124: There are a bunch of `400` case in this handler, how do you expect client apps to distinguish these errors?

src/controller/user.ts
```diff
@@ -0,0 +1,214 @@
...
+
+        if (!call.res) return response(context, 400, 'user_not_found')
+
+        call = await safeCall(
+            userRepository.findOne({ id: Not(Equal(userToBeUpdated.id)), email: userToBeUpdated.email }),
+        )
+
+        if (call.err) return response(context, 500, call.err)
+
+        if (call.res) return response(context, 400, 'user_already_exists')
```
haishanh, line 153: ```if (call.err) return response(context, 500, call.err)```

Any chance this can leak something we don't want to expose?

```
if (call.res) return response(context, 400, 'user_already_exists')
```

Here the response body is a string. Mixing content-type in API does not look like a good practice to me.

haishanh:
Regarding the refreshToken, I think I understand your points.

But extending the session with an expired accessToken looks wrong to me.

I think refreshToken should be returned to the client side (transport layer security should not be a concern anymore now days). If the client app is a browser one, it can opt to not save refreshToken in some unsafe storage or only have it in memory.

src/controller/user.ts
emmanuelnk: In the controller, along with the status code, the `request.body` returns a status message e.g. `user_not_found`, `user_already_exists` etc that can let the client know what wrong.
e.g.
```ts
if (call.res) return response(context, 400, 'user_already_exists')
```

src/controller/user.ts
emmanuelnk: At worst this will return an stringed Error message indicating database connection error since all format validation is handled before `safeCall` makes a database request and this function only wraps database calls. But you are right, it is better to not mix content-type and return `server_error` or `database_connection_error` depending on the needs of the receiving application.

emmanuelnk:
> But extending the session with an expired accessToken looks wrong to me.

I think it depends on the needs of the application and how the UI/UX works in the application. If you want to ensure _true sessions_ then yes, an expired token _should not allow a user to continue their session._ If you want a secure sliding session, then I think this method is a valid one (not necessarily the only one).   

Also, saving the `refreshToken` in memory in browser applications is safer than `sessionStorage` or `localStorage` (there's still some vulnerability to XSS attacks but not as bad) but it introduces one usability issue:
- The user looses the "session" if a user closes the browser tab. The in-memory variables are destroyed and thus the session cannot refresh and the user needs to log in again when they open a new tab despite not having closed the browser.

Ultimately, I'm not against sending `refreshTokens` to the browser and I'm not saying my implementation is the best one. It was more a design choice I made while trying to follow the recommendations of the OAuth2.0 specification.

xavierchow:
Thank you for review @haishanh @MiffyLiye,
It's good enough for our evaluation, I'm closing this PR.

@emmanuelnk we will get back to you soon.

PR #55, opened 2020-11-13
Test completed

ufo22940268:
I has done my home work. 

### The advance feature i implemented

-  [x] Imagine we have a new requirement right now that the user instances need to link to each other, i.e., a list of "followers/following" or "friends". Can you find out how you would design the model structure and what API you would build for querying or modifying it?
- [x] Related to the requirement above, suppose the address of user now includes a geographic coordinate(i.e., latitude and longitude), can you build an API that,
given a user name
return the nearby friends

### Scripts

- Start MongoDB
     
       docker-compose up -d

- Start server

      npm run serve

- Run tests 

      npm run tests

### API Docs

API Docs is in [wikis](https://github.com/ufo22940268/test-backend/wiki) of this project.


xavierchow:
thanks, closing.

PR #58, opened 2021-02-18
Dev

haluyhl:
Init PR with basic functionalities: general CRUD, follower API, tests
will continue to push docs, comments, and possibly some fixes(pm2 broken atm)


MiffyLiye:
Hi, I'm Tao from Wiredcraft, I will review the code and ask some questions.

README.md
```diff
@@ -1,73 +1,156 @@
...
+## Dependencies
+
+As I have not tested for version compatibilities, assume the latest version of following would work. 
+ - typescript/ts-node
+ - docker(for mysql and redis)
+ - jest/ts-jest
+ - pm2
+---
+## How
+   - docker run -d -p 6379:6379 --name redis1 redis
```
MiffyLiye, line 123: why not also put redis in docker-compose file?

haluyhl: could added a " redis: "  section in docker compose. 
I didn't plan to integrate this at first place, it was just a last minute 'lazy one liner'

MiffyLiye:
I found the web api always return status code 200 with error code in response body, which is not typical REST style, can you list the points that you think is essential for RESTful API design

haluyhl:
> I found the web api always return status code 200 with error code in response body, which is not typical REST style, can you list the points that you think is essential for RESTful API design

imo nowadays many so called restful apis are not really single-resource oriented. I think the essential part is to have the agreement across the team for an agreed standards of the design regard both business needs and dev needs. 
just my option tho, it might be me lacking of pro experiences.

src/lib/validator.ts
```diff
@@ -0,0 +1,44 @@
...
+    const ajv = new Ajv({})
+    const valid = ajv.compile(schema);
+    if (!valid(data)) {
+        return valid.errors![0];
+    } 
+}
+
+const userSchema = {
+  type: 'object',
+  properties: {
```
MiffyLiye, line 14: I found there are validation rules on the these fields, can I find the rules in the API document, API response (when validation fails), or test cases?

haluyhl: my api docs might be way too unprofessional. I think there is a validation case in test 'naively' demonstrates such check, in a rather simpler way, the request will end up with a predefined errCode. 

test('login json validates', async () => {
    const res = await server
        .post('/user')
        .send({
            userName: '123', // username contains unexpected combinations
            password: 'a', // min length violation
        })
    expect(res.body.errCode).not.toBe(-1)
})

src/services/user.ts
```diff
@@ -0,0 +1,232 @@
...
+      console.error(`Error updating password by username: ${err}`);
+      return userChangePasswordFail();
+    }
+
+  }
+  /**
+   * Get followers of a user
+   * @param username 
+   */
+  static async getFollowerByUsername(username: string): Promise<SuccessRes | ErrorRes> {
```
MiffyLiye, line 180: what will happen if one user has many followers (e.g. https://weibo.com/shanghaicity has 9 million followers)? how to solve the problem?

haluyhl: did you targeting the 'displaying follower' feature for this scenario? 
if so, the first thought that came across is to expand the request body for additional info like 
{
 where : {},
 sort: {},
 offset: 
 limit: 
} 
to limit the displaying to a certain range.

MiffyLiye: The query may exhaust database or web server's hardware resources without restrictions.
What index should be created if this api supports `where` & `sort` you mentioned above?

haluyhl: common 'where- ed' columns depend on the actual user stories.
might need some test ahead for Selectivity value > 0.9 (ideally) as an index. 
if not then combine where-ed columns as a composite index.

db change might happen as well, like split address to street, city, state, dob to y m d, but ideally would be to designed in advance with pre-planned user stories.

src/test/user.test.js
```diff
@@ -0,0 +1,112 @@
...
+const password = `p${Date.now()}`
+const testUser = {
+    name: userName,
+    password,
+    address: "1234 Main St",
+    description: "test description"
+}
+
+let COOKIE = ''
+
```
MiffyLiye, line 13: all the tests here only assert the `errCode` field, what benefit can we get from these tests?
there are many other styles of writing tests, e.g. test the function from the user's perspective / business perspective. 
can you compare the benefit & cost of different kinds of tests?

haluyhl: ideally would be testing against possible failure cases. the implementation above obviously is too naive just a showcase of test happening. 
testing against route level I/O would be simplest but harder to narrow down the actual cause of the failure, yet providing room for refactor in a different style fashion.
testing against actual execution logic/flow would be a rather detailed approach which makes locating problem easier but when refactor happens test cases would need to be updated as well, accordingly.

xavierchow:
Thanks @haluyhl @MiffyLiye , that's pretty much enough for our evaluation, 
I'm closing the PR.
We will get back to you soon @haluyhl

PR #60, opened 2021-03-02
test-backend

lqqyt2423:


MiffyLiye:
Hello, I'm Tao from Wiredcraft, I will review this PR and ask some questions.

README.md
```diff
@@ -19,55 +17,75 @@ Build a RESTful API that can `get/create/update/delete` user data from a persist
...
+#### Login
 
-## What We Care About
+```
+POST /login HTTP/1.1
+Content-Type: application/json
+{
+    "user": "admin",
+    "password": "123456"
+}
```
MiffyLiye, line 58: how to configure the user & password?

README.md
```diff
@@ -19,55 +17,75 @@ Build a RESTful API that can `get/create/update/delete` user data from a persist
...
 
-Feel free to use any open-source library as you see fit, but remember that we are evaluating your coding skills and problem solving skills.
+Response Body Like:
+{
+    "access_token": "xxx...",
+    "token_type": "bearer",
+    "expires_in": 7200
```
MiffyLiye, line 64: how to configure the `expires_in` time

MiffyLiye: and how to configure the jwt key

test/user.js
```diff
@@ -0,0 +1,125 @@
...
+        authorization = `Bearer ${res.body.access_token}`
+      })
+  })
+
+  afterEach(() => {
+    sinon.restore()
+  })
+
+  it('should GET /users 200', () => {
+    sinon.stub(User, 'find').resolves([])
```
MiffyLiye, line 26: Some teams think that trying to mock database is unwise/impractical for test cases, can you compare the benefit & cost of mocking database?

README.md
lqqyt2423: Could config in config.js, maybe get from environment variables in production.

test/user.js
lqqyt2423: Mocking is simple and useful in this scenario, no need to configure a test database, no dirty data will be generated.
When testing some complex functions, use a real database would be simple and convincing.

controllers/user.js
```diff
@@ -0,0 +1,96 @@
...
+  const user = await User.findById(id)
+  if (!user) {
+    ctx.status = 404
+    return
+  }
+  
+  user.name = payload.name
+  user.dob = payload.dob
+  user.address = payload.address
+  user.description = payload.description
```
xavierchow, line 69: just for example, what if there are a massive number of fields on the user model, 
would you still prefer to go with increasing code lines with `user.xxx = payload.xxx`?

lqqyt2423: `Object.keys(payload).forEach(k => { user[k] = payload[k] })`
or 
`UserModel.findByIdAndUpdate(user.id, { $set: { ...payload } })`

xavierchow: what if some of the properties of `payload` don't need to save into the user?

lqqyt2423: If have undefined properties, validator will throw error.

If have other defined properties and don't need to save, can write `user.xxx = payload.xxx` line by line, it is clearer.
Or just write code like `const { otherField, ...realPayload } = payload`.

xavierchow:
thanks, @lqqyt2423 
I'm going to close this PR and we will get back to you soon.

PR #61, opened 2021-03-02
Implemented restful API for Wiredcraft User Table

eronekogin:
## The restful API provides general functions as follows:

* Could create/list/delete/update a general user record by following the standard restful api format.
* Could create/list/delete/check a user's followings and followers.
* Could list/check a user's friends.
* Could list the user's nearby friends according to the user address's geo coordinates.
* Has a global logging mechanism which records every network access requests together with application internal events.

 This application is created and unit tested by Ian Jiang (mengyu.jiang@gmail.com).


myapp/README.md
```diff
@@ -0,0 +1,242 @@
...
+```
+* Make sure you could connect to the above configured redis server.
+* ```npm start```
+* Then you will see some similar messages as follows:
+```test
+> myapp@1.0.0 start D:\nodejs\workspace\test-backend\myapp
+> node ./bin/www
+
+D:\nodejs\workspace\test-backend\myapp\bin
+```
```
MiffyLiye, line 54: with `User.dbConfig.host: '192.168.0.107'`, the app cannot handle incoming request, 
are there other configurations that need to update to show the error details?

myapp/README.md
```diff
@@ -0,0 +1,242 @@
...
+| Link | ```http://localhost:3000/users/{userId}``` |
+| Method | DELETE |
+| Parameters | The input userId should be a positive integer |
+| Return | A json object will be returned to tell the user if the update is successful |
+
+---
+
+| Function | Create a single user |
+| :-- | :-- |
+| Link | ```http://localhost:3000/users/{userId}``` |
```
MiffyLiye, line 130: the link is not correct,
how to get the created user's userId?

MiffyLiye:
the http response status code is always 200, and has code 200, 201, 400, etc. in response body, which is not typical REST style, can you list the points that you think is essential for REST API design?

myapp/README.md
eronekogin: * As specified in the myapp/README.md file, the person who tries to deploy this application should make sure they are able to connect to the redis server specified by the host and port name.
* Currently the application does not provide any error message when we cannot connect to the remote redis server when starting up the application. If the database configuration is incorrect, it will hang.
* An enhancement for fixing this issue is to add a try/catch check when trying to create the redis client in the myapp/databases/redisClient.js module so that when the client creation is failing, an error will be sent back to the user console.

myapp/README.md
eronekogin: * Thank you so much for pointing out this typo in the README.md file, and it was created due to copy/paste from the description of other api.
* The actual design does not follow that description, and if the user access the home page of this application, he should be able to see the correct link for this user creation.

myapp/README.md
MiffyLiye: will the redis client retry to connect when there are network interruptions, 
what can we do to help detect failures and recover quickly in production environment?

eronekogin:
> the http response status code is always 200, and has code 200, 201, 400, etc. in response body, which is not typical REST style, can you list the points that you think is essential for REST API design?

* In my understanding, the standard RESTFUL API design pattern lives in an ideal world and it requires the agreements between the application and the user as follows:

### Standard REST API style:

* We should always use https protocol for the communications between the user and the api server.
* We should deploy the api under the specific domain like https://api.example.com.
* We should append the api's version number to the api's url like https://api.example.com/v1/users. Currently github follows this style.
* Each api url stands for a resource from the application and there should be no verb occurring in an api link.
* Then we use different HTTP methods to represent the actions we want to perform on the resource as follows:
  * GET: Fetch one or more instance of the target resource.
  * POST: Create a new resource.
  * PUT: Update an existing resource.
  * DELETE: Delete an existing resource.
* If the returned resources contain too many results, the api should provide filter parameters to the users and some normal parameters are as follows:
  * ?limit=10: Limit the amount of the returned resources to only 10 records.
  * ?page=2&pageSize=2: Paging the returned resources to contain only 2 records each page, then return the second page.
  * ?sortby=name&order=asc: Sorting the returned resources by the specified field name and order.
* The status code in the server response stands for the results from the api request:
  * 200 OK: Returned from GET method when the fetch is successful.
  * 201 CREATED: Returned from POST and PUT methods when the create/update is successful.
  * 204 NO CONTENT: Returned from DELETE method.
  * 400 INVALID REQUEST: Returned from POST/PUT methods when the delete is successful.
  * 401 Unauthorized: Returned when the user has no authorization to use this api.
  * 403 Forbidden: Returned when the user has been authorized but is not allowed to access this specific api.
  * 404 NOT FOUND: Returned when the user try to ask a resource which is not found on the server.
  * 422 Unprocesable entity: Returned from the POST/PUT methods when a validation error happens during create/update.
  * 500 INTERNAL SERVER ERROR: Returned when some application error happens on the server.
* If there is any error from the user request, the api should return error message as a json object in the body with error as key and the message text as value as follows:

```json
{
    "error": "Error message text"
}
```

* For the server returning values when the action is success:
  * GET /collection: Return an array of resource objects.
  * GET /collection/resource: Return the single resource object.
  * POST /collection: Return the created resource object.
  * PUT /collection/resource: Return the updated resource object.
  * DELETE /collection/resource: Return an empty object.

 * It is also suggested to follow HATEOAS style when creating a restful api.
 * The server response should usually be in json format.

### About the status code and response:

* In real practice, some communication company like China Telecom will hijack the response with non-200 status code, for example, 404 and redirect them to the other pages. This could also happen when using some unethical browsers. So I believe it is a bit safer to always return response code with status 200 and including the actual status code in the response object.
* So it is a strong suggestion to always use https protocols during communication between the users and the api server.
* Once the https communications are implemented for this application together with user authroization/authentication function, I would be glad to enhance the current api response to follow the standard style.

myapp/README.md
eronekogin: * The current version of this application is using node-redis^3.0.2, where the redis.createClient() method will create a client instance that will auto reconnecting to the specified redis server by default.
* For production environment, it is not good enough to just use the default parameters set by the redis.createClient() method but instead we should offer the retry_strategy with the specific requirement from the production environment, an example is as follows:

```javascript
const client = redis.createClient({
  retry_strategy: function(options) {
    if (options.error && options.error.code === "ECONNREFUSED") {
      // End reconnecting on a specific error and flush all commands with
      // a individual error
      return new Error("The server refused the connection");
    }
    if (options.total_retry_time > 1000 * 60) {
      // End reconnecting after a specific timeout and flush all commands
      // with a individual error, currently the limitation is 1 minute.
      return new Error("Retry time exhausted");
    }
    if (options.attempt > 10) {
      // End reconnecting with built in error, currently the limit is 10 times.
      return undefined;
    }
    // reconnect after a specific time.
    return Math.min(options.attempt * 100, 3000);
  },
});
```
  * The above strategy will prevent the application from holding up due to losing connection to the redis-server, and the customized error types could be caught by the application so that it could log the target error and send the response to the user without being frozen forever.
* Meanwhile, we could also encapsulate the client object as a proxy, which functions as follows:
  * When the actual redis client is not ready due to network broken or remote server error, the proxy object will have the same interfaces as a RedisClient object, but does no real work.
  * Once the actual redis client is ready, either initial connecting or recovered from disconnection, we set the proxy object to it and now the proxy object could do some real database work like sending commands to redis.

myapp/validates/UserValidation.js
```diff
@@ -0,0 +1,138 @@
...
+
+  /**
+   * Validate if the input latitude is between -85 to 85.
+   * @param {String} latitude 
+   * @returns {Boolean}
+   */
+  static isValidLatitude(latitude) {
+    if (VALID_GEO_COORDINATES.test(latitude)) {
+      const val = parseFloat(latitude);
+      if (val >= -85.0 && val <= 85.0) {
```
bsdelf, line 39: Does it mean people located at North/South Pole won't be able to use this application? https://www.esrl.noaa.gov/gmd/grad/antuv/SouthPole.jsp

myapp/tests/controllers/UserController.test.js
```diff
@@ -0,0 +1,1412 @@
...
+      ['HSET', addressKey, ...Object.entries(address).flat()],
+      ['GEOADD', GEO_SORTED_SET, address.longtitude, address.latitude, userKey]
+    );
+  }
+
+  commands.push(['SET', USER_HASH_PREFIX, numberOfRecords]);
+  await redisClient.batch(commands);
+};
+
+const cleanDB = async () => {
```
bsdelf, line 52: Perhaps use `FLUSHALL` for `cleanDB` could be more reliable and easy to maintain.

myapp/validates/UserValidation.js
eronekogin: * Since the current version of the application is taking advantages of GEOADD and GEOSERACH commands in redis to input the user's geo coordinates and calculate his nearby friends, there is some limitation for GEOADD command as follows:

> The command takes arguments in the standard format x,y so the longitude must be specified before the latitude. There are limits to the coordinates that can be indexed: areas very near to the poles are not indexable.

> The exact limits, as specified by EPSG:900913 / EPSG:3785 / OSGEO:41001 are the following:

> Valid longitudes are from -180 to 180 degrees.
> Valid latitudes are from -85.05112878 to 85.05112878 degrees.
> The command will report an error when the user attempts to index coordinates outside the specified ranges.  

* So that was the reason why the -85.0 and 85.0 validation for the latitude came in the first place. Unfortunately for the people around north/south poles, they cannot use the current version of this application :)

 ## Options to handle the coordinates near the north or the south pole:

This question does bring a good point on how to index for the coordinates near the north or the south pole. Here are a few options that I could think of:

* We could still use GEOADD/GEOSEARCH from redis to fetch our nearby friends when the latitude of those friends are not very large. According to the precision, the error rate from GEOHASH algorithm is as follows:
  * At the Equator (0 Degrees), the length of a longitude degree is 111.320 km, while a degree of latitude measures 110.574 km, an error of 0.67%.
  * At 30 Degrees (Mid Latitudes) the error is 110.852/96.486 = 14.89%.
  * At 60 Degrees (High Arctic) the error is 111.412/55.800 = 99.67%, reaching infinity at the poles.
* For edge cases like people with larger latitude, we could switch to PostGIS, which is a spatial database extended from postgreSql. It has rich support to spartial projecting and provides excellent functions to solve those edge cases.

myapp/tests/controllers/UserController.test.js
eronekogin: Whether to use FLUSHALL depends on the following cases:

* If no other developer is performing testing against the configured redis server, it is a good choice to use FLUSHALL command.
* If the redis server is shared by multiple people who could be performing their testing at the same time, I don't believe FLUSHALL command is a better idea.
* Another good practice is to provide value for the prefix parameter when creating the redis client. So that for each command it will append this prefix to the specified key before sending each command. So 'namespace:test' could be used as a prefix when executing those test cases.

myapp/README.md
MiffyLiye: Application that can automatically recover from short period interruptions will save a lot of DevOps effort.
When auto recovery becomes more and more complex, one disadvantage is that some edge cases are hard to test, we won't have confidence about what will happen in edge cases.
Another way to help recover is to use monitor, app can send health status to monitor (or let monitor pull health status from app), when app is unhealthy for some time, DevOps can receive alert and manually investigate & solve the issue.

xavierchow:
Thanks all, that's good enough for our evaluation, I'm gonna close this PR.

@eronekogin We will reach out to you soon.

PR #63, opened 2021-04-02
Submission from Kengo TODA

KengoTODA:
Hi folks,

I will submit my implementation and documents as follows. Please review when you have time :)
Picture and hyperlinks in docs probably don't work on the PR page, so use the following URLs to read them:

* [README.md](https://github.com/KengoTODA/test-backend/tree/c74a90ba25fb73b887423b16b9f2e0be878f2e46#readme)
* [ARCHITECTURE.md](https://github.com/KengoTODA/test-backend/blob/c74a90ba25fb73b887423b16b9f2e0be878f2e46/ARCHITECTURE.md)

[OpenAPI document and online demo](https://wiredcraft.kengo-toda.jp/docs/) are also available.

### Notes

- I assume that the API is for internal use and not designed to be published on the web. So I applied no auth, CSRF, CORS, and other security requirements. The `createdAt` prop is also updatable.
- I split the project into smaller tiers, which could be over-engineering for such a small project. It's intentional to tell what I care to code explicitly.


src/infra/on-memory/user.repository.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    const list = Array.from(this.map.values());
+    list.sort((a, b) => a.id.localeCompare(b.id));
+    let start = 0;
+    if (from) {
+      for (const user of list) {
+        if (from.localeCompare(user.id) < 0) {
+          break;
+        }
+        ++start;
```
xavierchow, line 29: can it be simplified with `Array.findIndex()`?

src/presentation/user.dto.ts
```diff
@@ -0,0 +1,98 @@
...
+  IsDateString,
+  IsNotEmpty,
+  IsNumberString,
+  IsOptional,
+  IsString,
+  Length,
+} from 'class-validator';
+import { ApiProperty } from '@nestjs/swagger';
+
+export class UserDto {
```
xavierchow, line 19: It seems the `UserDto` has the exact data shape with [NewUser](https://github.com/Wiredcraft/test-backend/pull/63/files#diff-a5105e8e39b26a4054abe7d84a51b24b045afb9789a26d03623f84313fdbc35cR12) except the decoration with class-validator.
Any idea to avoid duplication?

README.md
```diff
@@ -1,73 +1,124 @@
...
 
-### Advanced requirements
+```typescript
+private readonly friends: Set<UserId>;
+private readonly followers: Set<UserId>;
 
-*These are used for some further challenges. You can safely skip them if you are not asked to do any, but feel free to try out.*
+follow(another: User) {
+    this.follower.add(another.userId);
+    another.friends.add(this.userId);
```
xavierchow, line 71: This is a kind of bi-direction link, any idea to mitigate the potential data inconsistency? e.g. the `follower.add succeeds but the `friends.add` fails`?

README.md
```diff
@@ -1,73 +1,124 @@
...
-- Write clear documentation on how it's designed and how to run the code.
-- Write good in-code comments.
-- Write good commit messages.
-- An online demo is always welcome.
+When code, The 'follow' is `User`'s behaviour, so implement a `follow(User)` method in `User` class like:
 
-### Advanced requirements
+```typescript
+private readonly friends: Set<UserId>;
+private readonly followers: Set<UserId>;
```
xavierchow, line 67: If we have a need to get followers of a specific user with the order by `followed time`, how do you manage to design the data structure for `followers`?

src/presentation/user.dto.ts
KengoTODA: It's intentionally separated.

DTO and Entity have different responsibilities and belong to different tiers. If we mix them, controllers can use business logic in an entity like `follow(User)`, and domain models depend on things required from the presentation tier like validation. By separating them, we can make the responsibility of classes clear, and keep the code base safe to change.

README.md
KengoTODA: As described in [the case 1 in README](https://github.com/KengoTODA/test-backend/blob/c74a90ba25fb73b887423b16b9f2e0be878f2e46/README.md#case-1-friends-feature), [the event-driven approach](https://docs.microsoft.com/en-us/azure/architecture/guide/architecture-styles/event-driven) is one idea. I'm a fan of [the reactive manifecto](https://www.reactivemanifesto.org/) 😀. 

The transaction is acceptable but not the best for maintainability.

README.md
KengoTODA: The easiest idea is to replace this `Set` with `List`, and add new friends/followers to the end of these lists.
It would be enough if we just care about its order. We need less change even on the datastore side.

If we need to persist the `followed time` record, we can use `List<{userId: UserId, date: Date}>` in the domain tier.  
To design its data model for mongo, we have several ideas:

1. [Embedded data model](https://docs.mongodb.com/manual/core/data-model-design/#embedded-data-models), if and only if followers are enough little and we're sure that document size is smaller than [the maximum BSON document size](https://docs.mongodb.com/manual/reference/limits/#mongodb-limit-BSON-Document-Size).
2. [Normalized data model](https://docs.mongodb.com/manual/core/data-model-design/#normalized-data-models) with an isolated document that persists necessary props such as `user`, `follower`, and `created time`. The [compound index](https://docs.mongodb.com/manual/core/index-compound/) helps us to find documents by `user` and sort by `created time`.

KengoTODA:
Hi @xavierchow do you have any other questions? If not, I will shut down the server.

xavierchow:
@KengoTODA thank you for the reminder.
I have no further questions and you can safely shutdown your server.
Sorry for the late response, there are still some internal discussions about your application.
Please bear with us for a bit further more.

I'm gonna close the PR.

PR #65, opened 2021-04-15
Additional test - Obed

oseifrimpong:
## Refresh Mechanism
I have added redis for token management. All tokens have a ttl of 2 hours. 

Basically, when the `refresh` endpoint is called, the `refresh_token` will be verified and deleted from `redis` together with it's `access_token`. 

A new set of tokens will be generated and saved in redis.

## Logout
Accepts access token, verifies it and removes it from redis.

### To Improve
* Refactor authentication tests.
* Add a sequence diagram to represent the implementation.
* Improve on token invalidation

_I was not able to implement the points above due to work and time constraint._


xavierchow:
Thanks for your PR, I will start to review it soon.

src/services/auth-service.ts
```diff
@@ -0,0 +1,128 @@
...
+   * Takes a users and returns access and refresh tokens. 
+   * It also saves tokens into redis with ttl of 2 hours
+   * @param userP User Profile
+   * @returns 
+   */
+  async createTokens(userP: UserProfile): Promise<Tokens> {
+    let accToken = '';
+    let refToken = '';
+
+    const userId = String(userP[securityId])
```
xavierchow, line 40: any reason to use `String` class for coercion here?

src/services/auth-service.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    try {
+      accToken = await this.jwtService.generateToken(userP)
+      refToken = await JWTService.prototype.generateRefreshToken(userP)
+
+
+      // Save tokens in redis
+      if (accToken && refToken) {
```
xavierchow, line 48: is it necessary to have the existence check in the `if`?
I think an error will be thrown if there is any token generation failure and the `if` won't be reached at all, no?

src/services/auth-service.ts
```diff
@@ -0,0 +1,128 @@
...
+ * @param token a refreshed token
+ * @returns A new set of tokens.
+ */
+  async refreshToken(token: string): Promise<Tokens> {
+    if(!token){
+      this.logger.error('auth-service-refresh: token is missing');
+      throw new HttpErrors.BadRequest('Error refreshing token');
+  }
+    try {
+      const user = await JWTService.prototype.verifyRefreshToken(token)
```
xavierchow, line 91: any reason to use prototype? what's the difference from `this.jwtService.verifyRefreshToken`?

src/services/auth-service.ts
```diff
@@ -0,0 +1,128 @@
...
+      if (!user) {
+        this.logger.error('auth-service-refresh: Failed to verify token');
+        throw new HttpErrors.Unauthorized('Error refreshing token');
+      }
+      const tokens = await this.createTokens(user)
+      return tokens
+    } catch (error) {
+      this.logger.error('auth-service-refresh: ', error);
```
xavierchow, line 99: it seems there is inconsistency about the coding style(semicolon), do you have a lint set up for the code format?

src/services/auth-service.ts
```diff
@@ -0,0 +1,128 @@
...
+ * @returns a string 
+ */
+  async logout(token: string): Promise<String> {
+    if(!token){
+        this.logger.error('auth-service-logout: token is missing');
+        throw new HttpErrors.BadRequest('Error logging out');
+    }
+    try {
+      const user = await this.jwtService.verifyToken(token)
+      if (!user) {
```
xavierchow, line 116: I think `user` will never be nullable from the `verifyToken` function, no?

src/services/jwt-service.ts
```diff
@@ -0,0 +1,116 @@
...
+    return token;
+  }
+
+
+/**
+ * Takes a user's profile and generates a refresh Token
+ * @param userProfile 
+ * @returns a refresh token
+ */
+  async generateRefreshToken(userProfile: UserProfile): Promise<string> {
```
xavierchow, line 72: this function is almost the same as `generateToken` except for the error message and the secret.
There should be a space to improve to remove the duplicated codes.

xavierchow:
Two further questions,

# Refresh token
> when the refresh endpoint is called

when should the client(frontend application) call the refresh endpoint?

# Logout

> Accepts access token, verifies it and removes it from redis.

I see you remove the tokens from the redis when logging out, but I feel like both tokens are still available with consequent API calls, correct me if I'm wrong.

src/services/auth-service.ts
oseifrimpong: Yes, I do have lint setup but did not run it before submitting the code. I haven't set it up to run onSave listing. I usually have it set up as part of a `pre-commit` hook so it runs before any commit. Didn't consider adding that to this project.

oseifrimpong:
> Two further questions,
> 
> # Refresh token
> > when the refresh endpoint is called
> 
> when should the client(frontend application) call the refresh endpoint?
> 

- The client should call the refresh endpoint when the access_token expires. 
> # Logout
> > Accepts access token, verifies it and removes it from redis.
> 
> I see you remove the tokens from the redis when logging out, but I feel like both tokens are still available with consequent API calls, correct me if I'm wrong.

- Yes, you are right the tokens will be valid until it expires in Redis. I noticed this flaw and that is why I stated token invalidation as part of things to improve. Thought about the solution below but didn't have the time to implement it. 

The solution for this would be to have an interceptor to check the existence of a token in Redis and returning a `401` when token is not found in Redis.

src/services/auth-service.ts
oseifrimpong: Yes, `if` won't be reached when token generation fails, you are correct. I think it would be better to check for the  existence of previous tokens and removing them before pushing the new tokens. 
This way, if there is an error generating the token, the previous token will still be available in Redis until it expires.

src/services/auth-service.ts
oseifrimpong: Yes, this check is not necessary.

src/services/auth-service.ts
oseifrimpong: There is no particular reason actually, `userP[securityId].toString()` will also work fine, just that `toString()` will through an error when `userId` is null and `String` would not. But in this case, `userId` will not be null so it will be okay to use that as well.

src/services/auth-service.ts
oseifrimpong: There is no significant reason actually. It's just another way of accessing the method from the class. The `verifyRefreshToken` can be accessed by making it accessible via the constructor. 

The difference between the two is that `prototyping` allows you to access the method from the parent class without instantiating the class. While accessing the method via the constructor does the otherwise.

src/services/jwt-service.ts
oseifrimpong: Yes of course, there is definitely a way to improve this. This can be achieved by having a function that has both `user profile` and `secret` as parameters and have that implement the token signing.

xavierchow:
Thanks, that's good enough for our evaluation, closing this PR.

Will get back to you soon, thank you for your time and effort.

PR #67, opened 2021-04-28
Backend Test Submission by Feiran Yang 

myyukiho:
Hi, I spent a few nights learning Node.js and built a backend service which provides the basic functionalities of manipulating the user as required by @xavierchow , considering that I have no previous experience of Node.js/JavaScript.

### DIFF
--- a/src/configs/config.js
+++ b/src/configs/config.js
@@ -1,0 +1,30 @@
+// can add more configuration for prod environment
+const env = process.env.NODE_ENV; // 'dev' or 'test'
+
+const dev = {
+    app: {
+        port: parseInt(process.env.DEV_APP_PORT) || 3000
+    },
+    db: {
+        host: process.env.DEV_DB_HOST || 'localhost',
+        port: parseInt(process.env.DEV_DB_PORT) || 27017,
+        name: process.env.DEV_DB_NAME || 'dev'
+    }
+};
+const test = {
+    app: {
+        port: parseInt(process.env.TEST_APP_PORT) || 3000
+    },
+    db: {
+        host: process.env.TEST_DB_HOST || 'localhost',
+        port: parseInt(process.env.TEST_DB_PORT) || 27017,
+        name: process.env.TEST_DB_NAME || 'test'
+    }
+};
+
+const config = {
+    dev,
+    test
+};
+
+module.exports = config[env];
--- a/tests/routes/users.test.js
+++ b/tests/routes/users.test.js
@@ -1,0 +1,184 @@
+const app = require('../../src/index'); // link to the server file
+const supertest = require('supertest');
+const request = supertest(app);
+
+const mongoose = require('mongoose');
+const User = require('../../src/models/user');
+
+const config = require('../../src/configs/config');
+
+const { db: { host, port, name } } = config;
+const connectionString = `mongodb://${host}:${port}/${name}`;
+
+beforeAll(async () => {
+    // disconnect mongoose from the db of app instance
+    await mongoose.disconnect();
+    await mongoose.connect(connectionString, {
+        useNewUrlParser: true
+    });
+
+    const presetUsers = [{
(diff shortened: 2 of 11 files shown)


src/configs/config.js
```diff
@@ -0,0 +1,30 @@
...
+        host: process.env.TEST_DB_HOST || 'localhost',
+        port: parseInt(process.env.TEST_DB_PORT) || 27017,
+        name: process.env.TEST_DB_NAME || 'test'
+    }
+};
+
+const config = {
+    dev,
+    test
+};
```
MiffyLiye, line 28: when to use `dev` config, and when to use `test` config?
Image that this app will be deployed to staging environment, how to configure the app with environment variables?
And in staging environment, the database requires username / password authentication, how to configure the app?

tests/routes/users.test.js
```diff
@@ -0,0 +1,184 @@
...
+        const response = await request
+            .delete('/users/6088d0ccf5a4bb73ed24cfee');
+
+        expect(response.statusCode).toBe(404);
+        expect(response.body.message).toBe('User not found.');
+
+        done();
+    })
+
+    it('Should return 500 when trying to delete user by invalid ID', async done => {
```
MiffyLiye, line 175: can you explain why 500 status code is chosen in this case?
some people also suggest status code like 404, 200, 204 when ID is invalid, can you compare the pros and cons of these status codes?

src/configs/config.js
myyukiho: When you need some test data for local development, use `dev` config; when you need to run integration tests to check if the data actually hits the database, use `test` config.  

For staging environment, I would add a file which contains encrypted credentials to the machine hosting the staging environment, and add `staging` config to the app which reads username & password from the file.

tests/routes/users.test.js
myyukiho: I think it's just a typo here. In my opinion using 404 makes more sense because it indicates the failure of operation explicitly, however 200 and 204 are commonly used then the operation succeeds.

src/configs/config.js
MiffyLiye: this question is programming language independent, and database independent, 
the connection string in `app.js` doesn't read username / password, 
https://github.com/Wiredcraft/test-backend/blob/848b1d715d1673c2a4072ea01a4d4d842e9b7a15/src/app.js#L8-L9
I think simply replace `config.js` file won't work.

tests/routes/users.test.js
```diff
@@ -0,0 +1,184 @@
...
+    for (const u of presetUsers) {
+        const user = new User(u);
+        await user.save();
+    }
+})
+
+
+afterAll(async () => {
+    // clean up DB.
+    await User.deleteMany();
```
MiffyLiye, line 51: some people clean up DB before tests run, can you compare these two approaches?
also some people use `beforeEach` not `beforeAll` to do DB cleanup, can you compare these two approaches?

src/configs/config.js
myyukiho: I see. Regardless of programming language and database, I think the best practice is to put sensitive data in the file such as `db.yaml` on the host machine, and when app starts, it reads everything from the file based on different environments.  If using environment variables, for this app, it should be something like:
```
const { db: { username, password, host, port, name } } = config; 
const connectionString = `mongodb://${username}:${password}${host}:${port}/${name}`; 
```

tests/routes/users.test.js
myyukiho: I think you're right, cleaning up DB before tests run makes more sense because it could potentially clean up test data from previous tests left in the DB. 

Well, if using `beforeAll`, you need to clean up DB in the individual test, if there is some operations on DB in the test. Whereas using `beforeEach` can be less efficient than `beforeAll` because it cleans up and sets up DB more frequently, but it's also less error-prone.

xavierchow:
thanks all, closing.

PR #78, opened 2021-09-29
YuQi commit for backend test

burningyu:


src/modules/user/controller.js
```diff
@@ -0,0 +1,270 @@
...
+	startsWith, 
+} from 'lodash';
+import { User } from '../models';
+
+export async function create(ctx, next) {
+	let data = null;
+	let error = null;
+    let transaction;
+	try {
+        transaction = await service.transaction();
```
xavierchow, line 17: the indent is broken.

package.json
```diff
@@ -0,0 +1,99 @@
...
+    "koa-router": "^7.3.0",
+    "lodash": "^4.17.15",
+    "lru-cache": "^5.1.1",
+    "md5-file": "^5.0.0",
+    "moment": "^2.20.1",
+    "moment-timezone": "^0.5.33",
+    "mysql2": "^1.5.1",
+    "nedb": "^1.8.0",
+    "node-forge": "^0.7.5",
+    "node-rdkafka": "^2.7.4",
```
xavierchow, line 59: where is rdkafka used?

src/modules/core/db.logger.js
```diff
@@ -0,0 +1,54 @@
+const {
```
xavierchow, line 1: what is the difference between this file and src/modules/core/database.logger.js?

src/modules/user/controller.js
bsdelf, line 17: Transactions are used in `create`, `update`, `remove`, is it necessary? What is the underlying SQL statements in Sequelize `User.create`?

burningyu: Fix  on commit ID 7e918

package.json
burningyu: used for common message-center module and is not used currently.

src/modules/core/db.logger.js
burningyu: They are used for different db instance,currently are all for MySQL and no difference

src/modules/user/controller.js
burningyu: It’s necessary because these actions would modify data
User is import from arc/modules/models/user, and in this file return a MySQL model insatance using sequelize.define() function.So u can use MySQL model inner function or any other defined static function in this file from any where when imported it

PR #79, opened 2021-11-07
feat: implement backend

test3207:
### host

https://wiredcraft.test3207.com

### oauth login testing [here](https://wiredcraft.test3207.com/login)
* sometimes it doesn't work out because i used a server in mainland, which has connection problems to github, while other oauth provider in china require tons of paper. you can just try later.

### crud tests as following

PS: you can login from oauth service above, then you will be able to test with your own cookie.

```shell=
curl "https://wiredcraft.test3207.com/users?id=1" --compressed
```

```shell=
curl -X POST "https://wiredcraft.test3207.com/users" -H "cookie: backend=d89b4df275da8bc7f7587cebf4fcdaf4aa89f459c45e3fb47761647617af90df" --compressed
```
```shell=
curl -X PUT "https://wiredcraft.test3207.com/users?id=1" -H "cookie: backend=d89b4df275da8bc7f7587cebf4fcdaf4aa89f459c45e3fb47761647617af90df" -D "{"address":"random"}" --compressed
```
```shell=
curl -X DELETE "https://wiredcraft.test3207.com/users?id=1" -H "cookie: backend=d89b4df275da8bc7f7587cebf4fcdaf4aa89f459c45e3fb47761647617af90df" --compressed
```
please do not run this delete command. you'd better delete someone else's account.

### oauth

as documented in comment, an independent service is hosting for this purpose. codes are [here](https://github.com/test3207/oauth).

### logging

nah just debug for local developing. and collect all logs for production, delete logs that older than one year.

### follower/following

two seperate tables: follower and following.

and an extra column in user table named followers.

record all users one's following, set some sort of limit to allow users follow limited users when they follow to many people(like a captcha).

record say 1000 followers, in the followers table. if people keep follow/unfollow this guy for more than 1000 people, just calculate in user.followers, delete old records to make room for new followers.

### nearby

if a plugin is allowed, use postgis directly.
or set idx for lat, long, then search people in a square, finally narrow down the range(or not. no one really care lol)


xavierchow:
Hi @test3207 
What's the consideration of self-implemented API routes instead of frameworks like `express`, `koa`, etc.?

test3207:
> Hi @test3207 What's the consideration of self-implemented API routes instead of frameworks like `express`, `koa`, etc.?

relying on the framework will make people become the shape of the framework [doge].

i have been thinking about what would i do if it was me to design the structure of a project, so here comes my try.

frameworks dont really care about data sources mapping. you need to define it manully both in node and database. i hope my try in this merge somehow can show my thinking to solve it.

docs/user.md
```diff
@@ -0,0 +1,98 @@
...
+* get user info
+
+### Parameters
+
+|Name   |Type   |Optional   |In	Description |
+|-------|-------|-----------|---------------|
+|id     |numbser|t          |user id        |
+|name   |string |t          |user name      |
+
+### Response
```
bsdelf, line 18: What would the server return, if following errors occur?
- data not found
- internal error (for example: db connection lost)

src/service/session.ts
```diff
@@ -0,0 +1,97 @@
...
+    return Math.floor((expireAt.getTime() - Date.now()) / 1000);
+}
+export class SessionService {
+    private cookieName = 'backend';
+    constructor(private app:App) {
+    }
+    private sessionKey = (session:string) => {
+        return `backend:user:session:${session}`;
+    };
+    private sha256Encode = (data:string, salt:string) => {
```
bsdelf, line 17: Two questions:

- why is salt needed
- why use the same salt for all sessions

docs/user.md
test3207: * data not found returns  statuscode 404,
* not handling db connection at all, deamon will keep restarting service. in a higher level, nginx return 500.

src/service/session.ts
test3207: Theoretically its a wrong design, attackers can always generate enough rainbow tables offline to break things.

A fixed salt can only help a little by blocking newbiee hackers. But yes, it's wrong.

Not using bcrypt because it's slow, and cookie is verified too often. i only use bcrypt for passwords.

Some user-related salt can help. (salts updated in cron job).

bsdelf:
> please do not run this delete command. you'd better delete someone else's account.

So all registered users have access permission to all API right?

bsdelf:
@test3207 Do you have any idea about how to test:

- route, API
- OAuth module

test3207:
> > please do not run this delete command. you'd better delete someone else's account.
> 
> So all registered users have access permission to all API right?

yes. furthering access permission design not mentioned in request so i skipped it.

test3207:
> @test3207 Do you have any idea about how to test:
> 
> * route, API
> * OAuth module

routers still have some problems here, didnt parse search params carefully, and an extra feature not implemented, which is a wrapper that shape types for each api automatically, so i didnt start the router/api test. ideally it would be a similar way with service, a function called, assert the data. it may takes a few more days so....

for oauth module, it's a third party service so we can't actually ensure they are always functional. for codes in this project, i would fake some data as returns; for codes in that oauth repo, i may consider use pupeteer to run the real tests (with some extra accounts for test only)

xavierchow:
thanks all, I think the PR and disucssions here are good enough for our evaluation.
I'm gonna close this PR.
@test3207 we will get back to you soon.

PR #86, opened 2022-01-29
Jack liang.chen/dev

jack-liang-chen:
CRUD features implemented.

### DIFF
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -1,45 +1,69 @@
-import { v4 as uuidv4 } from "uuid";
+import { User } from "../models/user.js";
 
 let users = [];
 
 export const getUsers = (req, res) => {
-  res.send(users);
+  User.find({}, (err, data) => {
+    if (!err) {
+      res.send(data);
+    } else {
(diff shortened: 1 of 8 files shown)


controllers/users.js
```diff
@@ -0,0 +1,69 @@
...
+  });
+};
+
+export const createUser = (req, res) => {
+  const user = new User({
+    ...req.body,
+    createdAt: Date.now(),
+  });
+
+  user.save((err, data) => {
```
xavierchow, line 31: `err` is ignored?

controllers/users.js
```diff
@@ -0,0 +1,69 @@
...
+
+export const deleteUser = (req, res) => {
+  User.findByIdAndDelete(req.params.id, (err, data) => {
+    if (!err) {
+      res.status(200).json({
+        code: 200,
+        message: "User deleted successfully",
+        deletedUser: data,
+      });
+    }
```
xavierchow, line 67: if there is an `err`, the response hangs forever.

PR #88, opened 2022-03-06
my homework

dongsw95:


app/extend/customTransport.js
```diff
@@ -0,0 +1,41 @@
...
+        if (parseInt(i, 10) === args.length - 1) {
+          args[i] += '\n';
+        }
+      }
+    }
+
+    super.log(level, args, meta);
+  }
+
+  buildFormat(level) {
```
xavierchow, line 26: I don't see the param `level` is used in this function.

app/extend/helper.js
```diff
@@ -0,0 +1,7 @@
+const CryptoJS = require('crypto-js');
+
+module.exports = {
+  md5Encrypto(str) {
+    return CryptoJS.MD5(str).toString().toUpperCase();
```
bsdelf: The "crypto" package of nodejs already has md5 hash, what's the benefit of "crypto-js"?

app/extend/customTransport.js
```diff
@@ -0,0 +1,41 @@
...
+class CustomTransport extends FileTransport {
+  constructor(options, ctx) {
+    super(options);
+    this.ctx = ctx;
+  }
+
+  log(level, args, meta) {
+    const prefixStr = this.buildFormat(level);
+    for (let i in args) {
+      if (args.hasOwnProperty(i)) {
```
bsdelf: The type of `args` is `any[]` https://github.com/eggjs/egg-logger/blob/master/index.d.ts#L197, so why not use `for...of` statement to iterate over `args`?

app/controller/user.js
```diff
@@ -0,0 +1,144 @@
+'use strict';
+
+const Controller = require('egg').Controller;
+const moment = require('moment');
```
bsdelf, line 4: "moment" is not in package.json, seems it's a indirect dependency. BTW, this package is not recommended anymore, the moment official suggests to use "luxon" or "day.js" instead: https://momentjs.com/docs/#/-project-status/

app/service/user.js
```diff
@@ -0,0 +1,105 @@
...
+    });
+  }
+
+  /**
+   * updateUserById
+   * @param {int} id
+   * @param {obj} userOpt
+   * @returns User
+   */
+  async updateUserById(id, userOpt) {
```
bsdelf, line 64: what about the `pwd` field, will it be plain text after update?

app/controller/friend.js
```diff
@@ -0,0 +1,168 @@
...
+    if (id === userId) {
+      throw new Error('Can not follow yourself');
+    }
+    const followerInfo = await ctx.service.user.getUserById(id);
+    if (!followerInfo) {
+      throw new Error('The Person is not exist');
+    }
+    const followRelation = await ctx.service.friend.checkFollowRelation(userId, id);
+    console.log('relation:', followRelation);
+    if (followRelation & (1 === 1)) {
```
bsdelf: Seems it should be `(followRelation & 1) === 1`.

app/controller/friend.js
```diff
@@ -0,0 +1,168 @@
...
+    });
+  }
+
+  async removeFollower() {
+    const { ctx } = this;
+    const userId = ctx.headers.userId;
+    const id = +ctx.params.id;
+
+    const followRelation = await ctx.service.friend.checkFollowRelation(userId, id);
+    if (followRelation & (1 !== 1)) {
```
bsdelf: Similar issue as L19. Actually I think `checkFollowRelation` is not a good idea, because the data might be changed after query due to concurrency request, if user send two follow requests at the same time, both of them will probably pass the check and result in two records in database.

Better solution:

- add unique index `(following, follower)`
- insert new relationship into friend table and catch `SequelizeUniqueConstraintError`
- use return value of `destroy` to check whether the relationship has been removed: https://sequelize.org/v7/class/src/model.js~Model.html#static-method-destroy

dongsw95: Thanks forreviewing my homework. I hadn't realize so many bugs or shortcomings in it. And I had just committed for debugging.
- Using crypto instead of crypto-js.
- debug with plain text of pwd when updating user.
- debug with (followRelation & 1) === 1, that may caused by auto prettier :-(.

btw, I used for...in to get index of array, to add 'prefix' to args[0] and '\n' to args[args.length - 1], should i use for... instead?
for moment.js, it is a hlelpful package i used before, i will replace it since it is  in maintenance mode.
Then for multiple click of same request, i think unique index + transaction is a solution, but i didn't use it in this demo, i can do it this evening.

bsdelf: > btw, I used for...in to get index of array, to add 'prefix' to args[0] and '\n' to args[args.length - 1], should i use for... instead?

If you only want to change the first and the last item, why not directly access to `args[0]` and `args[args.length-1]` without loop? If you do want to mutate each array item, I think either `for (let i = 0; i < args.length; i++)` or `args.map()`is more readable.

dongsw95: > If you only want to change the first and the last item, why not directly access to `args[0]` and `args[args.length-1]` without loop? If you do want to mutate each array item, I think either `for (let i = 0; i < args.length; i++)` or `args.map()`is more readable.

roger that, thanks

xavierchow:
@dongsw95 thanks for your PR, I noticed that you organized your API test with postman collection which is fine.
However it only covers the public APIs, could you try to add some test codes with `jest` or `mocha`, I think either of them is the de-factor test framework for the node.js project in the industry.

dongsw95:
> @dongsw95 thanks for your PR, I noticed that you organized your API test with postman collection which is fine. However it only covers the public APIs, could you try to add some test codes with `jest` or `mocha`, I think either of them is the de-factor test framework for the node.js project in the industry.

Sure. As i said in README.md, add test file is in TODO list, i'll add some test cases tonignt.

dongsw95:
> @dongsw95 thanks for your PR, I noticed that you organized your API test with postman collection which is fine. However it only covers the public APIs, could you try to add some test codes with `jest` or `mocha`, I think either of them is the de-factor test framework for the node.js project in the industry.

Sure. As i said in README.md, add test file is in TODO list, i'll add some test cases tonignt.

xavierchow:
thank you, this should be good enough for our evaluation. closing now.

PR #93, opened 2022-04-10
APIs for get/create/update/delete user data

dd201702:
Please refer to docs/how-to-run.md for how I designed and how to run the project.

### DIFF
--- a/common/models/user.js
+++ b/common/models/user.js
@@ -1,0 +1,121 @@
+"use strict";
+const UserService = require("../services/user.service");
+
+module.exports = function (user) {
+  user.updateById = async (id, name, dob, address, description) => {
+    try {
+      return await UserService.updateById(id, name, dob, address, description);
+    } catch (err) {
+      return Promise.reject(err);
+    }
+  };
+  user.remoteMethod("updateById", {
+    description: "Update user by id.",
+    accepts: [
+      {
+        arg: "id",
+        type: "string",
+        required: true,
+        http: { source: "path" },
+        description: "user id",
+      },
+      { arg: "name", type: "string", description: "user name" },
+      {
+        arg: "dob",
+        type: "date",
+        description: "date of birth",
+      },
+      {
+        arg: "address",
+        type: "string",
+        description: "user address",
+      },
+      {
+        arg: "description",
+        type: "string",
+        description: "user description",
+      },
+    ],
+    returns: { arg: "result", type: "user", root: true },
+    http: { path: "/:id", verb: "patch" },
+  });
+
+  user.getById = async (id) => {
+    try {
+      return await UserService.getById(id);
+    } catch (err) {
+      return Promise.reject(err);
+    }
+  };
+  user.remoteMethod("getById", {
+    description: "Get user by id.",
+    accepts: [
+      {
+        arg: "id",
+        type: "string",
+        required: true,
+        http: { source: "path" },
+        description: "user id",
+      },
+    ],
+    returns: { arg: "result", type: "user", root: true },
+    http: { path: "/:id", verb: "get" },
+  });
+
+  user.getAll = async () => {
+    try {
+      return await UserService.getAll();
+    } catch (err) {
+      return Promise.reject(err);
+    }
+  };
+  user.remoteMethod("getAll", {
+    description: "Get all users. Result format: {size: n, data: [user]}",
+    accepts: [],
+    returns: { arg: "result", type: "user", root: true },
+    http: { path: "/", verb: "get" },
+  });
+
+  user.createUser = async (name, dob, address, description) => {
+    try {
+      return await UserService.createUser(name, dob, address, description);
+    } catch (err) {
+      return Promise.reject(err);
+    }
+  };
+  user.remoteMethod("createUser", {
+    description: "Create new user.",
+    accepts: [
+      { arg: "name", type: "string", required: true, description: "user name" },
+      {
+        arg: "dob",
+        type: "date",
+        required: true,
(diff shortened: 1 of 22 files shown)


MiffyLiye:
Hello, I'm Tao from Wiredcraft, I'm going to review the PR and ask some questions.
1. In test cases, finding user with non existed id got 200 status code, which is not typical RESTful API style. Can you compare the pros and cons of typical RESTful status code, and always 200 status code?
2. The test cases depends on test data prepared in previous test cases, making the test cases not independent. The F.I.R.S.T. principles suggest test cases to be `Isolated/Independent`. Can you compare benefit and cost of writing dependent and independent tests? Are there ways to reduce the cost in each case?

dd201702:
> Hello, I'm Tao from Wiredcraft, I'm going to review the PR and ask some questions.
> 
>     1. In test cases, finding user with non existed id got 200 status code, which is not typical RESTful API style. Can you compare the pros and cons of typical RESTful status code, and always 200 status code?
> 
>     2. The test cases depends on test data prepared in previous test cases, making the test cases not independent. The F.I.R.S.T. principles suggest test cases to be `Isolated/Independent`. Can you compare benefit and cost of writing dependent and independent tests? Are there ways to reduce the cost in each case?

Hello Tao, thank you for give me feedback.
For question 1, I agree with you that non existed id should not get 200 status code. So I update function code to return response as below format. I did similar changes for update by id. And update related unit tests.
```
{
  "statusCode": 404,
  "message": "Not Found"
}
```
For question 2, my unit tests are not `Isolated/Independent`, should use beforeEach and afterEach to prepare and clean db data for each unit test. My current test cases are easily to verify functions can be step by step in one time, but bad if some one case failed the rest cases will be affected. But sorry I don't have time to fix my code today. If possible I will try to fix it in several days.

common/models/user.js
```diff
@@ -0,0 +1,121 @@
+"use strict";
+const UserService = require("../services/user.service");
+
+module.exports = function (user) {
+  user.updateById = async (id, name, dob, address, description) => {
+    try {
+      return await UserService.updateById(id, name, dob, address, description);
+    } catch (err) {
+      return Promise.reject(err);
```
xavierchow: Wrapping error with `Promise.reject` is not necessary?

common/services/user.service.js
```diff
@@ -0,0 +1,124 @@
...
+    if (!moment(user.dob).isSame(dob, "day")) {
+      updateValues.dob = dob;
+    }
+    if (user.address != address) {
+      updateValues.address = address;
+    }
+    if (user.description != description) {
+      updateValues.description = description;
+    }
+    if (JSON.stringify(updateValues) === "{}") {
```
xavierchow: the validation in this function with so many `if` is tedious, 
could you find another way to make code more concise?

MiffyLiye:
> 
> Hello Tao, thank you for give me feedback. For question 1, I agree with you that non existed id should not get 200 status code. So I update function code to return response as below format. I did similar changes for update by id. And update related unit tests.
> 
> ```
> {
>   "statusCode": 404,
>   "message": "Not Found"
> }
> ```
> 
> For question 2, my unit tests are not `Isolated/Independent`, should use beforeEach and afterEach to prepare and clean db data for each unit test. My current test cases are easily to verify functions can be step by step in one time, but bad if some one case failed the rest cases will be affected. But sorry I don't have time to fix my code today. If possible I will try to fix it in several days.

you can also document the new strategy in comments, not necessary to implement.

common/models/user.js
dd201702: Hello Xavier, I just update code to update this kind of process.

common/services/user.service.js
dd201702: Hello Xavier, I just update code to refine the validation in this function.

dd201702:
> 



> you can also document the new strategy in comments, not necessary to implement.

Got you. 
To make my unit tests be `Isolated/Independent`,  could use current mongo db or mock mongo db, then use before/beforeEach to prepare test data and after/afterEach to clean test data.

dd201702:
> > Hello Tao, thank you for give me feedback. For question 1, I agree with you that non existed id should not get 200 status code. So I update function code to return response as below format. I did similar changes for update by id. And update related unit tests.
> > ```
> > {
> >   "statusCode": 404,
> >   "message": "Not Found"
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > For question 2, my unit tests are not `Isolated/Independent`, should use beforeEach and afterEach to prepare and clean db data for each unit test. My current test cases are easily to verify functions can be step by step in one time, but bad if some one case failed the rest cases will be affected. But sorry I don't have time to fix my code today. If possible I will try to fix it in several days.
> 
> you can also document the new strategy in comments, not necessary to implement.

I just add several unit tests, using before and after to make them Isolated/Independent.

common/models/user.js
```diff
@@ -1,12 +1,13 @@
...
 const UserService = require("../services/user.service");
+const logger = require("../utils/logger.js");
 
 module.exports = function (user) {
   user.updateById = async (id, name, dob, address, description) => {
     try {
       return await UserService.updateById(id, name, dob, address, description);
     } catch (err) {
-      return Promise.reject(err);
+      logger.error(err);
```
xavierchow, line 10: well...
This is ignoring the error thrown. @dd201702

xavierchow:
Thanks all, this PR is good enough for our evaluation.

@dd201702 we will get back to you soon.

PR #94, opened 2022-04-10
[冯春博] feat(server): 用户后端及接口文档

FengBryan:


MiffyLiye:
Hello, I'm Tao from Wiredcraft, I will review the code and ask some questions.
1. The API document and test cases didn't mention error cases, especially client side error cases. In typical REST API, what should be the response for error cases, can you give some examples?
2. In test cases, the register API uses `userName` field name, while login API uses `username` field name, while in API document, they both use `userName`. Can you suggest how to know if API document is up to date?
3. How do you arrange the test case's Given-When-Then in the test code? What type of information should be asserted in Then part?
4. In test cases, register API and login API seems to reuse same `userId`, but doesn't use same password, and login can also get 201 status code. Can you find the root cause why password verification is skipped?

src/util/deps.ts
```diff
@@ -0,0 +1,52 @@
...
+  deps: Array<Type<unknown> | string | symbol>,
+  func: (...args: Array<string | symbol | any>) => MaybePromise<RESULT>,
+) {
+  const logger = useLogger('System')
+  const exec = (ctx: INestApplicationContext) => {
+    const moduleRef = ctx.get(ModuleRef)
+    return func(...deps.map((dep) => moduleRef.get(dep, { strict: false })))
+  }
+
+  if (app) return exec(app)
```
xavierchow, line 35: This is interesting, could you explain what's the case that codes run after this line?

I saw the `app` is set from main.ts.

xavierchow, line 35: This is interesting, could you explain what's the case that codes run after this line?

I saw the `app` is set from main.ts.

FengBryan: First, when the application has been setup, use `setGlobalApplicationContext` to keep single context in it. Then u will have some situations to run some code with `runWithDeps`, this line means if `app` exists, exec the callback code immediately, or push it into `buffering` queue to exec

xavierchow: just curious about what's the use case for `or push it into buffering queue to exec`,
as the `app` is set from the main.ts, it should exist, no?

FengBryan:
@MiffyLiye I will patch a commit for issues

MiffyLiye:
> @MiffyLiye I will patch a commit for issues

it's OK to propose a plan or analyze the root cause  in comments, not necessary to implement.

src/util/deps.ts
FengBryan: For design, the `runWithDeps` function used to collection todos and do callback if app ready, it will not limit to run with clearly `timeline`. We can use runWithDeps anywhere, but execute when app ready. In some actual case to consider app state is hard, but make a queue is easy.

FengBryan: > For design, the `runWithDeps` function used to collection todos and do callback if app ready, it will not limit to run with clearly `timeline`. We can use runWithDeps anywhere, but execute when app ready. In some actual case to consider app state is hard, but make a queue is easy.

For this design, we can use `runWithDeps` in any lifecycle step of nestjs, such as `OnApplicatiionBootstrap`

FengBryan:
@MiffyLiye PTAL

MiffyLiye:
@FengBryan The implementation looks OK.
Can you choose one open question below and provide some thoughts?
Just share some thoughts, doesn't need to be perfect.

* Can you suggest how to know if API document is up to date?
* Can you find the root cause why password verification is skipped? How to help other team members avoid this in the future?

FengBryan:
> @FengBryan The implementation looks OK. Can you choose one open question below and provide some thoughts? Just share some thoughts, doesn't need to be perfect.
> 
> * Can you suggest how to know if API document is up to date?
> * Can you find the root cause why password verification is skipped? How to help other team members avoid this in the future?


**Can you suggest how to know if API document is up to date?**
1. With code change, decorator will update metadata to its controller class, and api server will get newest metadata when use `Reflect`, so the swagger doc will be up to date. This is automatic, no need to change manually
2. If there will be a new release and will not compatible for previous api define, will should reword api path with new version such as `v1 => v2`. If previous version will be deprecated, should note the doc item to deprecate

**Can you find the root cause why password verification is skipped? How to help other team members avoid this in the future?**

1. This is a low-level mistake i made, because i use `bcrypt` module for password validate, but i forgot it behavior. I thought it will throw error when compared fail. But it returned a boolean to sign success or fail. I just use compare but missed the return value, so it will login successfully when given invalid user info.
2. To avoid this problem, I think the best way is more and more test cases to cover all branches/scenarios.

xavierchow:
Thanks guys, very thorough discussion here, pretty good enough for our evaluation.
I'm closing this PR, we will get back to you soon @FengBryan

PR #95, opened 2022-04-17
User service with restful APIs

ShawnQing:
APIs, logging, test cases, swagger, documentation


src/service/user-service.js
```diff
@@ -0,0 +1,161 @@
...
+   * @return {{ docs: Array, total: number }}
+   */
+  getUser: async function(query, option, logger = log) {
+    const _log = logger.child({ module: 'user-service', method: 'getUser' });
+    _log.info('retrieve user, query: %j, option: %j', query, option);
+
+    const dbConnection = mongoSvc.getMongoConnection(_log);
+    if (dbConnection == null) {
+      _log.error('Cannot get mongodb connection.');
+      throw new errors.InternalServerError('Cannot get mongodb connection.');
```
xavierchow, line 41: The mongodb connection error handling seems to appear in many places in this user-service,
an improvement is expected.

test/user-router-test.js
```diff
@@ -0,0 +1,130 @@
...
+            delete mockData.id;
+            chaiHttpAgent.post(userRouterURI).send(mockData).end((err, res) => {
+                chai.expect(res).to.have.status(400);
+                done();
+            })
+        });
+    });
+    describe("#Get", function () {
+        it("Get User - Success - By ID", function (done) {
+            chaiHttpAgent.get(userRouterURI).query({id: userMock.id}).end((err, res) => {
```
xavierchow, line 48: Does it mean the `Get User` test cases are relying on the previous `Create User` test?

src/service/user-service.js
```diff
@@ -0,0 +1,161 @@
...
+   * @param body {{ id: String, name: String, dob: String, address: String, description: String }}
+   * @param logger { Pino }
+   * @return {{ docs: Array }}
+   */
+  createUser: async function(body, logger = log) {
+    const _log = logger.child({ module: 'user-service', method: 'createUser' });
+    _log.info('create user: %j', body);
+
+    const user = _.pick(body, ['id', 'name', 'dob', 'address', 'description']);
+    _.checkRequired(user, ['id', 'name']);
```
bsdelf, line 69: is any reason to use lodash and hand written code for object validation instead of use mature library like joi/ajv/...?

src/service/user-service.js
```diff
@@ -0,0 +1,161 @@
...
+   */
+  updateUser: async function(id, body, logger = log) {
+    const _log = logger.child({ module: 'user-service', method: 'updateUser' });
+    _log.info('update user, id: %s, body: %j', id, body);
+
+    if (_.isEmpty(id)) {
+      _log.error('id is required.');
+      throw new errors.BadRequestError('id is required.');
+    }
+    const user = _.pick(body, ['name', 'dob', 'address', 'description']);
```
bsdelf, line 105: `name` is required on creation, but it seems update operation doesn't have such check.

src/service/user-service.js
ShawnQing: Sure. I will improve it in next commit. Thanks  for your suggestion.

test/user-router-test.js
ShawnQing: Yes. The test cases are in the order `create -> get -> update -> delete` to make sure the mock data is reusable and will be cleaned at last, even though I am using memory mongo for testing.

src/service/user-service.js
ShawnQing: Thanks for your comments.

In my design, `id ` is the primary key for user model, user can update any other field by id, so `name ` is not required on update operation. Any concern, please let me know.

ShawnQing: If there is no `name` field in request body, name will not be updated in this operation.

src/service/user-service.js
ShawnQing: No special reason, haha.

I just thought this way is cleaner only for the field mandatory checking, no need a schema definition.
Of course, if we have the requirement of value validation, e.g. number checking, length checking, using joi library is a better choice.

Thanks for your suggestion. I will keep on enhancing my application.

src/service/user-service.js
bsdelf: I mean the create api forbids `name: null` while update api allows `name: null`

package.json
```diff
@@ -0,0 +1,42 @@
...
+    "chai-http": "^4.3.0",
+    "cookie-parser": "^1.4.6",
+    "cors": "^2.8.5",
+    "dotenv": "^16.0.0",
+    "express": "^4.17.3",
+    "helmet": "^5.0.2",
+    "jsonwebtoken": "^8.5.1",
+    "lodash": "^4.17.21",
+    "mocha": "^9.2.2",
+    "mocha-prepare": "^0.1.0",
```
bsdelf, line 25: "mocha" and "chai" can be dev dependencies.

xavierchow:
Thanks for the PR, closing.

@ShawnQing we will get back to you soon.

PR #98, opened 2022-06-26
test review request

Lellansin:
Hi there, glad to meet you. 

It's the phasic work waiting for check, and here's the [README](https://lellansin.github.io/test-backend/) for taking a look first.

Ask me if there are any questions.


src/service/account.ts
```diff
@@ -0,0 +1,70 @@
...
+   */
+  async signIn(email: string, password: string) {
+    // 1. Find user
+    const user = await this.userModel.getOneByEmail(email);
+    assert(!!user, ERROR.SERVICE_ACCOUNT_SIGNIN_NOTFOUND_EMAIL);
+
+    // 2. Verify password
+    if (user.password !== this.encodePassword(user, password)) {
+      // Could count+1 for count > 2 block the sign in
+      return Promise.reject(ERROR.SERVICE_ACCOUNT_SIGNIN_PASSWORD);
```
bsdelf, line 38: why not `throw ERROR.SERVICE_ACCOUNT_SIGNIN_PASSWORD` here, is it a style preference?

src/entity/relation.ts
```diff
@@ -0,0 +1,38 @@
+/**
+ * # Relation Entity
+ *
+ * | property    | Type       | Default     |
+ * |-------------|------------|-------------|
+ * | _id         | ObjectID   |
+ * | fromId      | ObjectID   |
+ * | toId        | ObjectID   |
```
bsdelf, line 8: is it possible to have duplicate (fromId, toId) pairs in relation entities?

src/service/account.ts
Lellansin: Um, when I first wrote about that, it looked like:

```ts
assert(
  user.password === this.encodePassword(user, password),
  ERROR.SERVICE_ACCOUNT_SIGNIN_PASSWORD,
)
```

What came to me later, is that I tought there might be more requirements like  frequent failed events blocking, thus why I turned it into if-else block, and in my option, inside this if-code-block should be a Promise (might be return from another function), so I decided to use a `Promise.rejection` as a Promise placeholder.

src/middleware/stat.ts
```diff
@@ -0,0 +1,34 @@
...
+ * Check [index](../modules/middleware.html) for more middleware.
+ *
+ * @module
+ */
+import Koa from 'koa';
+import { Provide } from '../util/container';
+import { Middleware } from '../util/web';
+
+@Provide()
+export class StatCostTime {
```
xavierchow, line 25: What's the usage of this StatCostTime middleware?

test/api.test.ts
```diff
@@ -0,0 +1,448 @@
...
+        persons[idx].id = res.id;
+      });
+      return request
+        .get('/user/')
+        .expect(200)
+        .then((response) => {
+          const list: any[] = response.body;
+          list.map((res, idx) => {
+            assert(res.email === persons[idx].email);
+            persons[idx].id = res.id;
```
xavierchow, line 133: it seems those test cases have to be run in a certain order as they're coupled.
e.g. if you don't have the `GET /user` run first you don't have the `id` on the person entity.

src/entity/relation.ts
Lellansin: There might be risk which introduced by development further, but not for current version. 

I'm not familiar with MongoDB very well, and what I was thinking is that the absolute solution to avoid duplication is provided by DB itself, like the Mysql's composite unique index.

So I guessed there might be same alternative in MongoDB, if true, all I need is to config it in MongoDB CLI, if false, I must ensure it in coding. And I prefer the harder way in this test.

When the follow-relation build through RelationService, there is [duplication check](https://github.com/Lellansin/test-backend/blob/master/src/service/relation.ts#L47) and a [lock](https://github.com/Lellansin/test-backend/blob/master/src/service/relation.ts#L53) for blocking repeat write. It's not perfect solution but can deal with most concurrent situation.

However I did make mistak that I didn't leave TODO to check the ultimate solution (Is MongoDB support composite unique index), that's my fault.

MiffyLiye:
1. README file mentions that the project has `100% test coverage`, but some parameters in API document are not tested, e.g. `dob` field when create user. what do you think should be covered in tests?
2. API document mentions that `dob` field when create user should be number type, but string type also works. The `dob` field in response when get user is Date type, but actual response is string in ISO 8601 format. Can you suggest how to know if API document is up to date?
3. Create user API doesn't return created user's id, can you suggest how to get created user's id?
4. List users API supports page parameter, but doesn't support page size (or limit) parameter, why not also support limit or page size?
5. How to know if list users API returned all users, or should caller request next page?
6. Some APIs return HTML pages, some return json data, and some return both on different conditions. What is the use case of returning HTML page?
7. Some API requires authentication, and requires two cookies. What are the two cookies used for, why not combine them to one cookie?

src/middleware/stat.ts
Lellansin: For test coverage, I wanted to check if there can be two or more `@Guard` to decorate one mehtod ([implement](https://github.com/Lellansin/test-backend/blob/master/src/util/web.ts#L145)) and I test it [here](https://github.com/Lellansin/test-backend/blob/master/src/controller/user.ts#L114)

test/api.test.ts
Lellansin: Yep, I know the order while Mocha run them, so I did that for ease.

I considered the coverage as most important goal, if we start to pay attention to test speed, like using concurrency test with Jest, the data initialization should move to common util function, and generate everytime for each case with isolation. I'm sure I can do it when it is necessary.

src/entity/relation.ts
bsdelf: perhaps you can simply set `_id` as `${fromId}_${toId}`, then catch conflict error and throw bad request.

Lellansin:
Hi @MiffyLiye, thanks for your reading, here's my answers.

> 1. README file mentions that the project has `100% test coverage`, but some parameters in API document are not tested, e.g. `dob` field when create user. what do you think should be covered in tests?

It's code line coverage, not for every use case, if we need a 100% use case test coverage, we should have the 100% complete case list before. Since I didn't get the list, I chose code lines to be coverd.

> 2. API document mentions that `dob` field when create user should be number type, but string type also works. The `dob` field in response when get user is Date type, but actual response is string in ISO 8601 format. Can you suggest how to know if API document is up to date?

Introducing a CI task to ensure it's up to date.

> 3. Create user API doesn't return created user's id, can you suggest how to get created user's id?

In my opinion, I just thought when there is `POST /user` for user creating, it seems the `user` would like an `article` resource more than an `account` for auth. So I make creating user process implemented through `POST /account/signup`, sign up is pairing with sign in, It makes sense to get user's id after signing in. 

> 4. List users API supports page parameter, but doesn't support page size (or limit) parameter, why not also support limit or page size?

it's easy to support it, but if I did I also need to adjust corresponding tests. Considering the time cost and difficulty, bewteen undocumented details and important documented features (like Auth/Follow/...),  I prefer to pay more attention to the important part in a short time during this test.

> 5. How to know if list users API returned all users, or should caller request next page?

I wrote this test with GitHub, and I chose the same way which GitHub did on its [commit list](https://github.com/Lellansin/test-backend/commits/master).

There are two reason for me:

- If there is total count for pagination, this project's user will look like a administrator which want to manage the `users`. And what I got from our documentation, is prefer Twitter-like user, which indicated next page of friends instead of entire users pagination.
- it's easy, and the ease is beautiful (PS: I think the Github's commit list is more beautiful than some completed pagination page).

> 6. Some APIs return HTML pages, some return json data, and some return both on different conditions. What is the use case of returning HTML page?

Honestly, I use it for unit test, LOL. Actually, I usually separate them with the `/api` prefix for normal APIs. But in this test, I considered it as a pure API server at beginning, which I prefer using NGINX-like proxy to add the prefix.

With one moment, I found there could be a cool way to show the auth flow with puppeteer in the e2e test (let's check the [gif](https://raw.githubusercontent.com/Lellansin/test-backend/master/asserts/unittest.gif)🤠), I temporarily add these render page.

If that matters, I can turn to my usual way (add `/api` prefix for all APIs).

> 7. Some API requires authentication, and requires two cookies. What are the two cookies used for, why not combine them to one cookie?

Indeed, there are two cookies like `xxx:sess`/`xxx:sess.sig`, which the `.sig` one is the signature for the other. As I know it's a security mechanism of [cookies](https://github.com/pillarjs/cookies) which is depended by Koa. We can check it with the [link](https://github.com/pillarjs/cookies/blob/0.8.0/index.js#L111).

I don't recommend turnning off this security mechanism, which means use only one cookie  to save data. Because it will increase the risk of suffering the "Man-in-the-Middle" attack.

src/entity/relation.ts
Lellansin: That will make geo query more difficult. And `fromId`/`toId` is ObjectId (for better geo relation query), which is fixed in 24 chars, `${fromId}_${toId}` is beyond 24 chars that could not become a `_id` of ObjectId.

So, I think it will be a good idea if we use that form with Redis cache before MongoDB queries.

Lellansin: That will make geo query more difficult. And `fromId`/`toId` is ObjectId (for better geo relation query), which is fixed in 24 chars, `${fromId}_${toId}` is beyond 24 chars that could not become a `_id` of ObjectId.

So, I think it will be a good idea if we use that form with Redis cache before MongoDB queries.

MiffyLiye:
> > 2. API document mentions that `dob` field when create user should be number type, but string type also works. The `dob` field in response when get user is Date type, but actual response is string in ISO 8601 format. Can you suggest how to know if API document is up to date?
> 
> Introducing a CI task to ensure it's up to date.
> 

Can you clarify what the CI task will do? 
Since the API document is no executable as unit tests, or postman scripts, how to know it matches the real implementation?

> > 5. How to know if list users API returned all users, or should caller request next page?
> 
> I wrote this test with GitHub, and I chose the same way which GitHub did on its [commit list](https://github.com/Lellansin/test-backend/commits/master).
> 
> There are two reason for me:
> 
> * If there is total count for pagination, this project's user will look like a administrator which want to manage the `users`. And what I got from our documentation, is prefer Twitter-like user, which indicated next page of friends instead of entire users pagination.
> * it's easy, and the ease is beautiful (PS: I think the Github's commit list is more beautiful than some completed pagination page).
> 

You can see that in the GitHub commits page, the `Newer` button is disabled, and `Older` button is clickable on first page, the `Older` button is disabled, and `Newer` button is clickable on last page.
GitHub's API gives you some way to know whether there is newer page or older page. 
How can the caller know if there is next page or not with your API response?

> > 7. Some API requires authentication, and requires two cookies. What are the two cookies used for, why not combine them to one cookie?
> 
> Indeed, there are two cookies like `xxx:sess`/`xxx:sess.sig`, which the `.sig` one is the signature for the other. As I know it's a security mechanism of [cookies](https://github.com/pillarjs/cookies) which is depended by Koa. We can check it with the [link](https://github.com/pillarjs/cookies/blob/0.8.0/index.js#L111).
> 
> I don't recommend turnning off this security mechanism, which means use only one cookie to save data. Because it will increase the risk of suffering the "Man-in-the-Middle" attack.

Can you explain more about the security risk, how to attack if there is no sig cookie verification?

Can you compare your session implementation with the following approaches:
1. generate random session id as access token, and save session payload in backend, set the session id in cookie.
2. generate session payload and signature and combine them in one string as access token, set that string in cookie, backend doesn't save session payload.

Lellansin:
> Can you clarify what the CI task will do? 

CI task can be:

- non-invasive
  - Use AST parse to get it from the code (high-cost).
  - Design unit test, and get it from analysis of test results.
- invasive for old version
  - implements `@Query`, `@Body`, `@Header` decorators to register the parameters for doc.

And more:

- In runtime (follow observer pattern):
    - Get sampling with API from middleware to generate result sample and its doc. (This is part of what I've done during Alibaba, which is a feature called request redo for debug).

> Since the API document is no executable as unit tests, or postman scripts, how to know it matches the real implementation?

If we don't have playground, we should check If the API's version and document's matches. Or try the runtime solution mentioned above.

> You can see that in the GitHub commits page, the `Newer` button is disabled, and `Older` button is clickable on first page, the `Older` button is disabled, and `Newer` button is clickable on last page. GitHub's API gives you some way to know whether there is newer page or older page. How can the caller know if there is next page or not with your API response?

For the website to get next page, there were two selections for me:

First one:
<table>
<tr><th>Side</th><th>Content</th></tr>
<tr>
	<td>Backend</td>
	<td>
		Implements total count function for:
		user/follower/followings/nearby-normal/nearby-follower/nearby-followings, and update their tests.
	</td>
</tr>

<tr>
	<td>Frontend</td>
	<td><pre>
if (page >= Math.round(total / limit)) {
    // last page!
}</pre></td>
</tr>
</table>

Second one:
<table>
<tr><th>Side</th><th>Content</th></tr>
<tr>
	<td>Backend</td>
	<td>
		Do nothing more.
	</td>
</tr>

<tr>
	<td>Frontend</td>
	<td><pre>
if (resultList.lenght < limit) {
    // last page!
} else if (!await fetch(/* another page */)) {
    // last page!
}</pre></td>
</tr>
</table>

Thorough ROI analysis within this test, I prefer the 2nd selection.

> Can you explain more about the security risk, how to attack if there is no sig cookie verification?

I can use my mobile phone to put a WiFi hotspot in the *****, which take the same name as the public WiFi in the *****, like *****, and then modify DNS resolve on my mobile phone to my local HTTP proxy server. Then, when the user's request arrives at my proxy server, I can parse the user's request and tamper with the cookie. If the cookie is not signed, the server will consider the cookie as true and use it. (Updated: hide some sensitive words to `*****`.)

If you noticed HTTPS, I may say there are ways to downgrade it to HTTP for some websites, or for clareless users.

> Can you compare your session implementation with the following approaches:
>
> 1. generate random session id as access token, and save session payload in backend, set the session id in cookie.
>2. generate session payload and signature and combine them in one string as access token, set that string in cookie, backend doesn't save session payload.

1.  High performance, high cost.
2. The performance is poor, and sometimes the gateway will reject the request with HttpStatusCode like 431 while the cookie is too large (4KB/8KB/..., according to the gateway), but low cost.

xavierchow:
Thanks all, the PR and the discussions are pretty enough for our evaluation, closing it now.

We will get back to you soon. @Lellansin

